PR #9, opened 2017-12-19
Remove blaze in favour of setup.py

juliaogris:
This PR removes all blaze BUILD file and CI references.
It also introduces a restructure of the code (see `ls <repo>/src`)
Setup tools are now used for `build`, `install`, `test` and distribution (see README.rst)

TODO
* Add `entry_points` to `setup.py` so that sysl and reljam can be executed without `python -m`
* Add `lxml` to dependencies (there problems with required `STATIC_DEPS=true` in env)
* Give instructions for installing behind corporate proxy ;)


.gitignore
```diff
@@ -1 +1,6 @@
-/bazel-*
+*.pyc
+.cache
+sysl.egg-info
+dist
+build
+.eggs
```
anzdaddy: Minor quibble: all non-empty files should end in a newline. Ditto for other files below.

anzdaddy: (It's not evident in PR view, but you can see missing-newline indicators in files view.)

juliaogris: Totally agreed. I've already added a `.editorconfig` file. I wanted to wait for the next PR to run pylint (PEP 8 standard) and editorconfig over all source files.

PR #25, opened 2018-01-08
Annotations

juliaogris:
Fixes #24 .

Changes proposed in this pull request:
- Add syntax `*:` in addition to ` .. * <- * :` for annotations
- Remove pdb invocations
- Add unittests for annotations
- Add e2e (end-to-end) tests for annotations

@anz-bank/sysl-developers

PR #27, opened 2018-01-10
json serializer fixes

sahejsingh:
Fixes #32 
Changes proposed in this pull request:
- Implements JSON serializer and deserialzer for Tuple based model
This change adds the required methods in the generated Java Serialiser code.
The unit test shows how to use the generated java code to serializer and deserialize the model. 

@anz-bank/sysl-developers

PR #33, opened 2018-01-15
Move e2e output to <repo>/tmp

juliaogris:
- Move generated e2e output on `<repo>/tmp location` 
In accordance with [previous comment](https://github.com/anz-bank/sysl/pull/25#discussion_r160290819), temporary, generated code used in tests should not sit alongside "real" source code.
- Adjust `.gitignore` accordingly

@anz-bank/sysl-developers

### DIFF
--- a/test/e2e/test_e2e.py
+++ b/test/e2e/test_e2e.py
@@ -25,7 +25,8 @@ def test_e2e(fname):
     dname = path.dirname(__file__)
     root = path.join(dname, 'input')
     model = '/' + fname
-    out_fname = path.join(dname, 'actual_output', fname + '.txt')
+    out_fname = path.join(dname, '..', '..', 'tmp',
+        'test', 'e2e', 'actual_output', fname + '.txt')
     remove_file(out_fname)
 
     cmd = ['sysl', '--root', root, 'textpb', '-o', out_fname, model]
--- a/test/e2e/test_e2e.py
+++ b/test/e2e/test_e2e.py
@@ -2,7 +2,6 @@ from sysl.core import syslloader
 from sysl.util.file import filesAreIdentical
 
 import pytest
-import os
 
 from os import path, remove
 from subprocess import call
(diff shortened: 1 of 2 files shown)


juliaogris:
@camh-anz are you asking re `Remove linter warnings` commit or the whole PR?

Re `Remove linter warnings`: I was thinking of you as I pushed this 'fix-up' commit. I came to the conclusion rather than rebasing and force pushing to my public fork (am I missing something? is there no force push required?) I will stick to my current workflow, keep the commits and then squash-merge at the end. Please let me know if you have concerns.

Re the whole PR: this change relates to an [old PR comment](https://github.com/anz-bank/sysl/pull/25#discussion_r160290819) that temporary, generated code used in tests should not sit alongside "real" source code. As we are creating more test cases with generated code  (e.g. #27) this seems like the right move for repo hygiene. I've updated the PR description accordingly. Better?

camh-anz:
I was commenting on the commit linked to the comment : "move e2e tests..."

test/e2e/test_e2e.py
```diff
@@ -25,7 +25,8 @@ def test_e2e(fname):
     dname = path.dirname(__file__)
     root = path.join(dname, 'input')
     model = '/' + fname
-    out_fname = path.join(dname, 'actual_output', fname + '.txt')
+    out_fname = path.join(dname, '..', '..', 'tmp',
+                          'test', 'e2e', 'actual_output', fname + '.txt')
```
camh-anz: This is rather fragile. It will fail if this file is moved in the repository and this path is not adjusted. Perhaps it needs a TODO to come up with better tmpfile handling?

anzdaddy: My original suggestion was to put things in a temp directory (/tmp on *nix). <repo>/tmp was a compromise to keep everything in one place, but still close to the project.

It wasn't a big deal for me, so if it's going to cause more trouble than it's worth, I'm entirely happy for us to drop this PR leave things as they are.

camh-anz: The right way to do this would be to split dname into the repository root and subpath, and then insert "tmp" in the middle. That also removes the duplication of "test/e2e" as well. But there's no canonical way to identify the repository root. If this was something that needed to be done in many places, it would be worth doing right, but if n==1 then it can wait for the poor sucker who implements n>1 to do.

juliaogris: Right you are. I've updated this file as good as I could.

PR #37, opened 2018-01-29
Travis CI using OSX

juliaogris:
- OSX builds on Travis excluding Java test
- Updated documentation

@anz-bank/sysl-developers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,22 @@
 language: python
 python:
   - "2.7"
-branches:
-  only:
-  - /.*/
+matrix:
+  include:
+    - os: linux
+      dist: trusty
+    - os: osx
+      language: generic
 env:
   - SYSL_PLANTUML=http://www.plantuml.com/plantuml
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew list python &>/dev/null || brew install python; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export PATH=/usr/local/opt/python/libexec/bin:$PATH; fi
 install:
   - pip install . pytest flake8
 script:
   - flake8
   - pytest
-  - cd test/java && gradle test
+  - if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then cd test/java && gradle test; fi
+fast_finish: true
--- a/README.rst
+++ b/README.rst
@@ -17,9 +17,10 @@ The set of outputs is open-ended and allows for your own extensions. Sysl has be
 Installation
 ------------
 If you are interested in trying out Sysl, you will need to install `Python 2.7 <https://www.python.org/downloads/>`_ .
-OSX users, we recommend installing Python 2.7 with `Homebrew <https://brew.sh/>`_ ::
+OSX users, we recommend installing Python 2.7 with `Homebrew <https://brew.sh/>`_  and adding it to the `PATH` in you `.bash_profile`::
 
   > brew install python
+  > echo export PATH=/usr/local/opt/python/libexec/bin:\$PATH >> ~/.bash_profile
 
 You can then install Sysl with ::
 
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -17,9 +17,10 @@ The set of outputs is open-ended and allows for your own extensions. Sysl has be
 Installation
 ------------
 If you are interested in trying out Sysl, you will need to install `Python 2.7 <https://www.python.org/downloads/>`_ .
-OSX users, we recommend installing Python 2.7 with `Homebrew <https://brew.sh/>`_ ::
+OSX users, we recommend installing Python 2.7 with `Homebrew <https://brew.sh/>`_  and adding it to the `PATH` in you `.bash_profile`::
 
   > brew install python
+  > echo export PATH=/usr/local/opt/python/libexec/bin:\$PATH >> ~/.bash_profile
```
camh-anz: Why python/libexec/bin and not just python/bin? Usually you would not have a libexec directory in your path as it contains internal executable components of a package.

.travis.yml
```diff
@@ -1,14 +1,22 @@
...
+  include:
+    - os: linux
+      dist: trusty
+    - os: osx
+      language: generic
 env:
   - SYSL_PLANTUML=http://www.plantuml.com/plantuml
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew list python &>/dev/null || brew install python; fi
```
camh-anz: &> is a bash-ism and I cannot see anywhere here or in the travis docs that ensure these scripts are run with bash. It would be safer to do: >/dev/null 2>&1

.travis.yml
```diff
@@ -1,14 +1,22 @@
...
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export PATH=/usr/local/opt/python/libexec/bin:$PATH; fi
```
camh-anz: You should not put a libexec directory on your path generally. .../python/bin should be sufficient. If it's not, it deserves a comment as to why.

.travis.yml
```diff
@@ -1,14 +1,22 @@
...
+matrix:
+  include:
+    - os: linux
+      dist: trusty
+    - os: osx
+      language: generic
 env:
   - SYSL_PLANTUML=http://www.plantuml.com/plantuml
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
```
camh-anz: [[ ... ]] is a bashism. Without anything specifying bash, you should stick to posix shell constructs:

if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew update; fi

Note the single equals sign instead of double-equals (which is also a bash-ism)

.travis.yml
juliaogris: Brew suggests it as it's moving to a `python2` world:
```
> brew info python
==> Caveats
This formula installs a python2 executable to /usr/local/bin.
If you wish to have this formula's python executable in your PATH then add
the following to ~/.zshrc:
  export PATH="/usr/local/opt/python/libexec/bin:$PATH"

Pip and setuptools have been installed. To update them
  pip2 install --upgrade pip setuptools

You can install Python packages with
  pip2 install <package>

They will install into the site-package directory
  /usr/local/lib/python2.7/site-packages
```
https://github.com/Homebrew/homebrew-core/issues/15746

README.rst
juliaogris: See above

.travis.yml
juliaogris: Done

.travis.yml
juliaogris: Done. I'm learning new stuff - thanks!

.travis.yml
camh-anz: You missed the last line of my comment, re: = vs ==. Posix does not support ==, only =

juliaogris: Apologies! I've got great intentions to read reviews more carefully from here onwards.
Done.

.travis.yml
camh-anz: I did not look closely enough in my /usr/local/opt/python/bin directory - it does NOT contain any executable called "python", only "python2", so .../python/bin is not a substitute for the libexec/bin directory. So carry on, and ignore me.

PR #38, opened 2018-01-29
Fix seq diagram parameter parsing

### LINKED ISSUE #36: sd endpoint is a required argument
### Expected behaviour
Produce a sequence diagram using the demo bank model by just providing the application

### Actual behaviour
Error and stack trace when command line parsing gets to endpoints

```
PS C:\dev\sysl> sysl sd --app ATM --plantuml http://localhost:8080/plantuml /demo/bank/bank -o out\atm.txt
Traceback (most recent call last):
  File "C:\tools\winpython\2.7.12.4\python-2.7.12.amd64\Scripts\sysl-script.py", line 11, in <module>
    load_entry_point('sysl==0.1.1', 'console_scripts', 'sysl')()
  File "C:\tools\winpython\2.7.12.4\python-2.7.12.amd64\lib\site-packages\sysl\core\__main__.py", line 125, in main
    args.func(args)
  File "C:\tools\winpython\2.7.12.4\python-2.7.12.amd64\lib\site-packages\sysl\core\syslseqs.py", line 470, in cmd
    endpoints=[i[0] for i in args.endpoint],
TypeError: 'NoneType' object is not iterable
```

juliaogris:
Fixes #36 

- Fix parameter parsing for `sysl sd` to require specific endpoint or output pattern
- Improve UX: only show stack trace if `--trace` is specified

@anz-bank/sysl-developers

### DIFF
--- a/src/sysl/core/__main__.py
+++ b/src/sysl/core/__main__.py
@@ -102,6 +102,9 @@ def main():
     argp.add_argument('--version', '-v',
                       help='show version number (semver.org standard)',
                       action='version', version='%(prog)s ' + __version__)
+    argp.add_argument('--trace', '-t',
+                      help=argparse.SUPPRESS,
+                      action='store_true', default=False)
 
     subparser = argp.add_subparsers(help='\n'.join(['sub-commands',
                                                     'more help with: sysl <sub-command> --help', 'eg: sysl pb --help']))
@@ -121,8 +124,13 @@ def main():
         except OSError as exc:
             if exc.errno != errno.EEXIST or os.path.isdir(args.output):
                 raise
-
-    args.func(args)
+    try:
+        args.func(args)
+    except Exception as e:
+        if args.trace:
+            raise
+        else:
+            print e
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


src/sysl/core/__main__.py
```diff
@@ -121,8 +124,13 @@ def main():
...
                 raise
-
-    args.func(args)
+    try:
+        args.func(args)
+    except Exception as e:
+        if args.trace:
+            raise
+        else:
+            print e
```
camh-anz: Since you've added error trapping, I would also add sys.exit(1) here and a sys.exit(0) on the no-error exit path. This makes the tool more usable from scripts that want to detect errors.

juliaogris: Great advice, thanks, Done.
I don't think the `sys.exit(0)` is necessary - did you suggest it for symmetry reasons?

camh-anz: It's not strictly necessary. I like it because 1) it says "you got here, everything is good", and 2) I lose track of what language does what - e.g. shell will exit with the error code of the last run statement, and 3) "explicit is better than implicit".

But up to you.

juliaogris: Totally convinced.

PR #59, opened 2018-02-07
Fix release script and CI config

juliaogris:
Fixes 58# and 43# (hopefully).

- Update Appveyor encrypted OAuth token (fingers crossed).
- Improve release script style
- Add automatic deployment to PyPI from Travis on tags.

@anz-bank/sysl-developers

### DIFF
--- a/src/scripts/release.sh
+++ b/src/scripts/release.sh
@@ -23,7 +23,7 @@ fi
 
 ORIGIN_URL=$(git remote get-url origin)
 UPSTREAM="anz-bank"
-UPSTREAM_URL=$(echo "$ORIGIN_URL" | sed -n "s/\\(.*\\/\\/github.com\\/\\)\\(..*\\)\\(\\/..*.git\\)/\\1$UPSTREAM\\3/p")
+UPSTREAM_URL=$(echo "$ORIGIN_URL" | sed -n "s|\\(.*//github.com/\\)\\(..*\\)\\(/..*.git\\)|\\1$UPSTREAM\\3|p")
 
 echo "------- Checkout master ---------"
 git checkout master || fatal "Cannot checkout master"
@@ -31,12 +31,12 @@ git checkout master || fatal "Cannot checkout master"
 echo "------- Pull upstream ---------"
 git pull "$UPSTREAM_URL" master || fatal "Cannot pull  upstream master"
 
-if [[ $COMANND = "prepare" ]]; then
+if [[ $COMMAND = "prepare" ]]; then
     RELEASE_BRANCH="release-v$VERSION"
-    ORIGIN=$(echo "$ORIGIN_URL" | sed -n "s/.*\\/\\/github.com\\/\\(..*\\)\\/..*.git/\\1/p")
-    REPO=$(echo "$ORIGIN_URL" | sed -n "s/.*\\/\\/github.com\\/..*\\/\\(..*\\).git/\\1/p")
+    ORIGIN=$(echo "$ORIGIN_URL" | sed -n "s|.*//github.com/\\(..*\\)/..*.git|\\1|p")
+    REPO=$(echo "$ORIGIN_URL" | sed -n "s|.*//github.com/..*/\\(..*\\).git|\\1|p")
     echo "------- Create release branch $RELEASE_BRANCH ---------"
-    git co -b "$RELEASE_BRANCH" || fatal "Cannot create release branch $RELEASE_BRANCH"
+    git checkout -b "$RELEASE_BRANCH" || fatal "Cannot create release branch $RELEASE_BRANCH"
 
     echo "------- Update version and commit ---------"
     echo "__version__ = '$VERSION'" > 'src/sysl/__version__.py' || fatal "Could not override src/sysl/__version__.py"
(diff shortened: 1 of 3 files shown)


juliaogris:
Annoyingly something is broken with Appveyor creating Github releases for organisations. 
I've created an [issue](https://github.com/appveyor/ci/issues/2068) on the Appveyor issue tracker and the [forum](https://help.appveyor.com/discussions/problems/11865-deployment-to-github-releases-for-organisation-seems-broken)

src/scripts/release.sh
```diff
@@ -23,20 +23,20 @@ fi
 
 ORIGIN_URL=$(git remote get-url origin)
 UPSTREAM="anz-bank"
-UPSTREAM_URL=$(echo "$ORIGIN_URL" | sed -n "s/\\(.*\\/\\/github.com\\/\\)\\(..*\\)\\(\\/..*.git\\)/\\1$UPSTREAM\\3/p")
+UPSTREAM_URL=$(echo "$ORIGIN_URL" | sed -n "s|\\(.*//github.com/\\)\\(..*\\)\\(/..*.git\\)|\\1$UPSTREAM\\3|p")
```
camh-anz: My eyes are still bleeding, even with the slashes now non-escaped.

I suggest using single quotes and Extended Regular Expressions (EREs) with sed to make this cleaner (there's a missing escape for the dot-git, which will mean one backslash will survive, so single-quotes will stop this doubling-up):

`sed -E -n 's|(.*//github.com/)(.+)(/.+\.git)|\1'"${UPSTREAM}"'\3|p`

Maybe I'm missing something, but if $ORIGIN_URL is a single URL, you don't need `sed -n` with the `p` modifier.

src/scripts/release.sh
```diff
@@ -23,20 +23,20 @@ fi
...
 echo "------- Pull upstream ---------"
 git pull "$UPSTREAM_URL" master || fatal "Cannot pull  upstream master"
 
-if [[ $COMANND = "prepare" ]]; then
+if [[ $COMMAND = "prepare" ]]; then
     RELEASE_BRANCH="release-v$VERSION"
-    ORIGIN=$(echo "$ORIGIN_URL" | sed -n "s/.*\\/\\/github.com\\/\\(..*\\)\\/..*.git/\\1/p")
-    REPO=$(echo "$ORIGIN_URL" | sed -n "s/.*\\/\\/github.com\\/..*\\/\\(..*\\).git/\\1/p")
+    ORIGIN=$(echo "$ORIGIN_URL" | sed -n "s|.*//github.com/\\(..*\\)/..*.git|\\1|p")
+    REPO=$(echo "$ORIGIN_URL" | sed -n "s|.*//github.com/..*/\\(..*\\).git|\\1|p")
```
camh-anz: Some comment here re EREs, single quotes and `sed -n`

src/scripts/release.sh
juliaogris: Much nicer, thanks. Done. 
(I can't possibly be held responsible for bleeding eyes).

src/scripts/release.sh
juliaogris: Done.

src/scripts/release.sh
```diff
@@ -23,20 +23,20 @@ fi
 
 ORIGIN_URL=$(git remote get-url origin)
 UPSTREAM="anz-bank"
-UPSTREAM_URL=$(echo "$ORIGIN_URL" | sed -n "s/\\(.*\\/\\/github.com\\/\\)\\(..*\\)\\(\\/..*.git\\)/\\1$UPSTREAM\\3/p")
+UPSTREAM_URL=$(echo "$ORIGIN_URL" | sed -E 's|(.*//github.com/)(.*)(/.*\.git)|\1'"${UPSTREAM}"'\3|')
```
camh-anz: You may as well make this $UPSTREAM and not ${UPSTREAM} as the rest of the script uses the first form. I always use the second and I type it without thinking which is why I said it last time - it makes no difference most of the time.

Also, did `+` not work in the regex? I recall having issues with it in the past, but if you're not using `.+`, you should use `..*` as you were before.

juliaogris: Argh - sorry, just sloppy on my side. Something else didn't work blablabla
Done.

src/scripts/release.sh
```diff
@@ -23,20 +23,20 @@ fi
 
 ORIGIN_URL=$(git remote get-url origin)
 UPSTREAM="anz-bank"
-UPSTREAM_URL=$(echo "$ORIGIN_URL" | sed -n "s/\\(.*\\/\\/github.com\\/\\)\\(..*\\)\\(\\/..*.git\\)/\\1$UPSTREAM\\3/p")
+UPSTREAM_URL=$(echo "$ORIGIN_URL" | sed -E 's|(.*//github.com/)(.+)(/.+\.git)|\1'"$UPSTREAM"'\3|')
```
marcelocantos: Needs an explanatory comment

src/scripts/release.sh
```diff
@@ -23,20 +23,20 @@ fi
...
 
 echo "------- Pull upstream ---------"
 git pull "$UPSTREAM_URL" master || fatal "Cannot pull  upstream master"
 
-if [[ $COMANND = "prepare" ]]; then
+if [[ $COMMAND = "prepare" ]]; then
     RELEASE_BRANCH="release-v$VERSION"
-    ORIGIN=$(echo "$ORIGIN_URL" | sed -n "s/.*\\/\\/github.com\\/\\(..*\\)\\/..*.git/\\1/p")
-    REPO=$(echo "$ORIGIN_URL" | sed -n "s/.*\\/\\/github.com\\/..*\\/\\(..*\\).git/\\1/p")
+    ORIGIN=$(echo "$ORIGIN_URL" | sed -E 's|.*//github.com/(.+)/.+\.git|\1|')
```
marcelocantos: Ditto

src/scripts/release.sh
juliaogris: ? This is a merged PR, I've implemented the changes Cam suggested on chat. Do you want me to write that here? Functionality doesn't change, just less bleeding eyes for @camh-anz as he put it.

PR #73, opened 2018-02-14
Rework e2e tests to be included in coverage

### LINKED ISSUE #71: Codecoverage does not include e2e tests
### Expected behaviour
We expect `pytest test/e2e/test_e2e.py` coverage to work and report ~20%.

### Actual behaviour
However, it reports coverage of ~11% even if no `sysl` command is executed, and it reports 0 if there are no sys specific imports in the e2e-test file but the `sysl` command _is_ executed.
This is related to how coverage instrumentalises python files - execution has to happen in the same process (http://coverage.readthedocs.io/en/coverage-4.2/subprocess.html).

```
coverage erase
coverage run --source=src/sysl -m py.test test/e2e/test_e2e.py
coverage html && open ./htmlcov/index.html
```

juliaogris:
Fixes #71  .

- rework `sysl.main` to take argument as optional (default `sys.argv[1:]`) parameter for testing
- extend e2e tests to optionally directly call `sysl.main` rather than start a subprocess

@anz-bank/sysl-developers

### DIFF
--- a/test/e2e/test_e2e.py
+++ b/test/e2e/test_e2e.py
@@ -1,4 +1,4 @@
-from sysl.core import syslloader
+from sysl.core.__main__ import main
 from sysl.util.file import filesAreIdentical
 
 import pytest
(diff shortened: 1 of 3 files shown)


test/e2e/test_e2e.py
```diff
@@ -16,14 +16,19 @@ def remove_file(fname):
...
+    ('004_annotations', True),
+    ('005_annotations', True),
+    ('001_annotations', False),
+    ('002_annotations', False),
+    ('003_annotations', False),
+    ('004_annotations', False),
+    ('005_annotations', False)
 ])
-def test_e2e(fname):
+def test_e2e(fname, subprocess):
```
sahejsingh: why do you need to subprocess=True calls?

juliaogris: I suppose I want true "end-to-end" tests from an enduser perspective, who doesn't care wether the program was written in Python or anything else.
Eventually we might swap out the executables underneath but the tests should remain the same (at least the subprocess=True ones). 
It also covers cases where the command line operates unexpectedly different from calling `main` directly. 
Does any of this convince you?

sahejsingh: Yup. thanks!

PR #68, opened 2018-02-13
Automate Bintray upload with Gradle on Travis for Java sysl-lib

juliaogris:
Fixes #46 and #70 .

Changes proposed in this pull request:
- Add Windows.exe download instructions to installation section of README
- Add `gradle.build` to Java sysl-lib and move Java sysl-lib sources to `src/libs/java/src/main/java` (this looks confusing, but gradle requires the `<project_root>/src/main/java` structure and considers `<repo>/src/libs/java` as `<project_root>` in this setup.)
- Add deployment of Java sysl-lib to bintray/JCenter as part of gradle.build script and Travis config
- Update release.sh script to also override gradle.properties version number
- Update tests of Sysl generated java code gradle.build
- Fix py2exe package collection on Appveyor
- Fix link from Appveyor badge on README to point to `master`

@anz-bank/sysl-developers

PR #78, opened 2018-02-14
Bump version to 0.1.4

juliaogris:


src/libs/java/gradle.properties
```diff
@@ -1 +1 @@
-syslVersion=0.1.4-alpha-5
+syslVersion=0.1.4
```
juliaogris: Unfortunately I issued this PR before the one #68 with the script update to automate this was merged.
Hopefully this won't be necessary again in the future.

PR #117, opened 2018-02-28
Add java.sysl and relocate json.sysl

anzdaddy:


juliaogris:
Please issue PRs from your fork at https://github.com/anzdaddy/sysl into upstream [`master`]((https://github.com/anz-bank/sysl/) for the following reasons:
* We (the @anz-bank/sysl-developers) work the same way as any potential open source contributor and find difficulties that way (there are some with CI and security related configs)
* We don't have to think about branch naming conventions to avoid naming conflicts
* We keep the branches of `upstream`  clean.

sysl2/lib/lang/java.sysl
```diff
@@ -0,0 +1,889 @@
+# sysl.lang.java
```
juliaogris: This is somewhat of a duplicate from the docs: https://github.com/anz-bank/sysl/blob/master/docs/ideas/grammar.md - can you please delete the docs section and put a link to this file there?
It seems like the grammar's grammar is loosely based on 
https://github.com/anz-bank/sysl/blob/sysl2/sysl2/proto/grammar.proto (which hasn't been reviewed).
This seems to be quite a core concept, so can we please have some documentation on this grammar - the `proto` file is not the best for comprehension.
You might want to use the [grammar.md](https://github.com/anzdaddy/sysl/blob/master/docs/ideas/grammar.md) as a starting point

sysl2/lib/lang/java.sysl
```diff
@@ -0,0 +1,889 @@
...
+!grammar java:
+    Identifier:
+        IdentifierChars - (Keyword | BooleanLiteral | NullLiteral)
+
+    IdentifierChars:
+        /[[:letter:]][[:letter:][:digit:]]/
```
juliaogris: `:letter:` and `:digit:` aren't defined anywhere, right? Also, isn't there a `*` missing?

sysl2/lib/lang/java.sysl
anzdaddy: Agreed we need docs #123, but shouldn't block this PR.

juliaogris: Well, I'd prefer to have things nice, clean and complete - remember the [broken window theory](https://blog.codinghorror.com/the-broken-window-theory/) ?

Otherwise we run risk of having heaps of Issues for outstanding documentation and/or tests.
It doesn't need to be much, probably pulling out the comments from the `proto` file and/or updating [grammar.md](https://github.com/anzdaddy/sysl/blob/master/docs/ideas/grammar.md) would be sufficient and I bet it's done in less than 5 minutes. It would set a great example.

If you absolutely insist that this cannot be done now and/or is likely to change so much in the imminent future that it doesn't warrant the effort, please let me know.

juliaogris:
BTW, CI is borked. [This line](https://github.com/anz-bank/sysl/blob/master/test/e2e/test_e2e.py#L105) in e2e testing is the culprit - unrelated to this PR, it's my bad I forgot to delete it when updating the svg output tests in a haste. You can either fix it yourself or I can create another PR.

sysl2/lib/lang/java.sysl
marcelocantos: `[:digit:]` is a standard posix regex element https://goo.gl/F9zQKy. `[:letter:]` should have been `[:alpha:]`, but really this is just rough thinking at the moment. It's not intended as a final compilable grammar yet.

juliaogris: ok.

juliaogris:
As per our discussion
* let's not worry about docs for now 
* let's merge these branches into `master` and work with a   
  `origin:branch` -> `upstream:master` PR model from here onwards

sysl2/lib/lang/java.sysl
juliaogris: @anzdaddy - I'm still wondering about the missing `*`, but happy I'm happy for you to merge anyway.

juliaogris:
@anzdaddy - I'm still wondering about the missing *, but happy I'm happy for you to merge anyway.

PR #131, opened 2018-03-16
Sysl2 java grammar

sahejsingh:
- Initial version of Backtracking parser.
- Build First and Follow set to analyse grammars

@anz-bank/sysl-developers

### DIFF
--- a/sysl2/Grammar1_test.go
+++ b/sysl2/Grammar1_test.go
@@ -1,8 +1,9 @@
 package main
 
 import (
-    sysl "anz-bank/sysl/sysl2/proto"
     "testing"
+
+    sysl "anz-bank/sysl/sysl2/proto"
 )
 
 func testParser(g *sysl.Grammar, numTerms int, tokens []int, text string, expected bool, t *testing.T) {
@@ -33,29 +34,34 @@ func TestGrammar(t *testing.T) {
     text := "bab"
     testParser(makeGrammar1(), 4, tokens, text, true, t)
 }
+
 func TestGrammar1(t *testing.T) {
     tokens := []int{2, 1, 1, 1, 0, -1}
     text := "bcccd"
     testParser(makeGrammar1(), 4, tokens, text, true, t)
 }
+
 func TestGrammarIncompleteInput(t *testing.T) {
     tokens := []int{2, 3, -1}
     text := "ba"
 
     testParser(makeGrammar1(), 4, tokens, text, false, t)
 }
+
 func TestEXPR1(t *testing.T) {
     text := "1 + 3 * 7"
     tokens := []int{6, 0, 6, 2, 6, -1}
     testParser(makeEXPR(), 7, tokens, text, true, t)
 
 }
+
 func TestOBJ(t *testing.T) {
     text := "{}"
     tokens := []int{0, 2, -1}
 
     testParser(makeRepeatSeq(makeQuantifierZeroPlus()), 4, tokens, text, true, t)
 }
+
 func TestOBJ2(t *testing.T) {
     text := "{123}"
     tokens := []int{0, 1, 2, -1}
@@ -67,6 +73,7 @@ func TestOBJ3(t *testing.T) {
     tokens := []int{0, 1, 3, 1, 3, 1, 2, -1}
     testParser(makeRepeatSeq(makeQuantifierZeroPlus()), 4, tokens, text, true, t)
 }
+
 func TestOBJPlusNegative(t *testing.T) {
     text := "{123}"
     // NOTE THE +
@@ -77,6 +84,7 @@ func TestOBJPlusNegative(t *testing.T) {
     tokens := []int{0, 1, 2, -1}
     testParser(makeRepeatSeq(makeQuantifierOnePlus()), 4, tokens, text, false, t)
 }
+
 func TestOBJPlusPositive(t *testing.T) {
     text := "{123, 234}"
     tokens := []int{0, 1, 3, 1, 2, -1}
@@ -88,6 +96,7 @@ func TestOBJPlusPositive(t *testing.T) {
     //    ;
     testParser(makeRepeatSeq(makeQuantifierOnePlus()), 4, tokens, text, true, t)
 }
+
 func TestOBJOptional(t *testing.T) {
     text := "{}"
     tokens := []int{0, 2, -1}
@@ -98,6 +107,7 @@ func TestOBJOptional(t *testing.T) {
     //    ;
     testParser(makeRepeatSeq(makeQuantifierOptional()), 4, tokens, text, true, t)
 }
+
 func TestOBJOptional1(t *testing.T) {
     text := "{123}"
     tokens := []int{0, 1, 2, -1}
@@ -108,6 +118,7 @@ func TestOBJOptional1(t *testing.T) {
     //    ;
     testParser(makeRepeatSeq(makeQuantifierOptional()), 4, tokens, text, true, t)
 }
+
 func TestOBJOptional2(t *testing.T) {
     text := "{123, 234}"
     tokens := []int{0, 1, 3, 1, 2, -1}
@@ -118,6 +129,7 @@ func TestOBJOptional2(t *testing.T) {
     //    ;
     testParser(makeRepeatSeq(makeQuantifierOptional()), 4, tokens, text, true, t)
 }
+
 func TestOBJOptionalNegative(t *testing.T) {
     text := "{123, 234, 567}"
     tokens := []int{0, 1, 3, 1, 3, 1, 2, -1}
@@ -128,16 +140,19 @@ func TestOBJOptionalNegative(t *testing.T) {
     //    ;
     testParser(makeRepeatSeq(makeQuantifierOptional()), 4, tokens, text, false, t)
 }
+
 func TestJSON_1(t *testing.T) {
     text := "{}"
     tokens := []int{3, 4, -1}
     testParser(makeJSON(makeQuantifierZeroPlus()), 8, tokens, text, true, t)
 }
+
 func TestJSON_2(t *testing.T) {
     text := `{ "abc" : 123 }`
     tokens := []int{3, 5, 6, 7, 4, -1}
     testParser(makeJSON(makeQuantifierZeroPlus()), 8, tokens, text, true, t)
 }
+
 func TestJSON_3(t *testing.T) {
     text := `{
         "abc" : 123 ,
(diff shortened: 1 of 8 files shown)


sysl2/Lexer_test.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+    "testing"
+)
+
+func TestLexer(t *testing.T) {
+    content := `keywords:
+        while | return |
+        if | else | import`
+
+    regexes := make([]string, 7)
```
marcelocantos: Literal slice

sysl2/grammarParser.go
```diff
@@ -0,0 +1,225 @@
+package main
+
+import (
+    sysl "anz-bank/sysl/sysl2/proto"
+    "sort"
```
marcelocantos: Move "sort" first, with a space between the two.

sysl2/grammarParser.go
```diff
@@ -0,0 +1,225 @@
...
+                str = `\A(?:` + str + `)`
+            case *sysl.Atom_Choices:
+                b.handleChoice(t.GetAtom().GetChoices())
+            }
+            if str != "" {
+                _, has := b.tokens[str]
+                if !has {
+                    b.tokens[str] = b.index
+                    b.arr = append(b.arr, str)
+                    t.Atom.Id = b.index
```
marcelocantos: No need for `index`. Use `len(b.arr)` instead.

sysl2/grammarParser.go
```diff
@@ -0,0 +1,225 @@
...
+                    b.index++
+                }
+            }
+        }
+    }
+}
+
+// NOTE: assigns new value to Atom.Id
```
marcelocantos: No need to write "NOTE:". This is implicit in comments.

sysl2/util.go
```diff
@@ -0,0 +1,35 @@
+package main
+
+// Set is Not exported
+type Set map[int]bool
```
marcelocantos: Use `struct{}` instead of `bool`.

sysl2/util.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (s *Set) clone() *Set {
+    ss := make(Set)
+    for k := range *s {
+        ss[k] = true
+    }
+    return &ss
+}
+
+func (s *Set) minus(tok int) {
```
marcelocantos: Call it `remove`.

sysl2/sysl_parser.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// CheckGrammar returns true if the grammar consumes the whole string
+// g - grammar to use
+// text - to parse
+// Start rule
+func checkGrammar(g *sysl.Grammar, tokens []int, start string) bool {
+    result, out, tree := parse(g, tokens, g.Rules[start])
+    fmt.Println(tree)
+    return (result && len(out) == 0)
```
marcelocantos: Remove outer parens.

sysl2/grammarParser.go
```diff
@@ -0,0 +1,225 @@
...
+    }
+    updated = true
+    for updated {
+        updated = false
+        for ruleName, rule := range g.Rules {
+            for _, seq := range rule.Choices.Sequence {
+                for i := range seq.Term {
+                    // follow
+                    l := len(seq.Term)
+                    if (i + 1) < l {
```
marcelocantos: Remove parens.

sysl2/Grammar1_test.go
```diff
@@ -0,0 +1,212 @@
...
+    }
+}
+
+// S –> bab | bA
+// A –> d | cA
+func TestGrammar(t *testing.T) {
+    tokens := []int{2, 3, 2, -1}
+    text := "bab"
+    testParser(makeGrammar1(), 4, tokens, text, true, t)
+}
```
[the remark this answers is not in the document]
sahejsingh: vscode shows run/debug code snippet over func definitions. it seems there is a line between functions!

sysl2/Grammar1_test.go
```diff
@@ -0,0 +1,212 @@
...
+    testParser(makeJSON(makeQuantifierZeroPlus()), 8, tokens, text, true, t)
+}
+func TestJSON_2(t *testing.T) {
+    text := `{ "abc" : 123 }`
+    tokens := []int{3, 5, 6, 7, 4, -1}
+    testParser(makeJSON(makeQuantifierZeroPlus()), 8, tokens, text, true, t)
+}
+func TestJSON_3(t *testing.T) {
+    text := `{
+        "abc" : 123 ,
```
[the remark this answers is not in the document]
sahejsingh: that is not code :). this is input to the lexer.

sysl2/Lexer_test.go
sahejsingh: the indexes are used in writing the tokens array. I need it this way.

sysl2/grammarParser.go
sahejsingh: vscode did its auto thing here again..
forcibly changed the order now.

sysl2/grammarParser.go
sahejsingh: done.

sysl2/util.go
sahejsingh: curious as to why?

sysl2/util.go
sahejsingh: done.

sysl2/sysl_parser.go
sahejsingh: done.

sysl2/grammarParser.go
sahejsingh: done.

sysl2/grammarParser.go
sahejsingh: done.

sysl2/grammarParser.go
```diff
@@ -184,7 +182,7 @@ func buildFirstFollowSet(g *sysl.Grammar) (map[string]*Set, map[string]*Set) {
                 for i := range seq.Term {
                     // follow
                     l := len(seq.Term)
-                    if (i + 1) < l {
+                    if i+1 < l {
```
marcelocantos: `if i + 1 < l … else if i < l …` is a little confusing. The equivalent `if i < l - 1 … else if i == l - 1` is easier to grok. Additionally, you could use `last := len(seq.Term) - 1` instead of `l := len(seq.Term)`.

sysl2/sysl_parser.go
```diff
@@ -137,5 +137,5 @@ func parse(g *sysl.Grammar, input []int, val interface{}) (bool, []int, []interf
 func checkGrammar(g *sysl.Grammar, tokens []int, start string) bool {
```
marcelocantos: Rename to `grammarIsValid` or some similar predicate form — anything that makes grammatical sense (excuse the pun) in an `if` statement.

sysl2/util.go
```diff
@@ -1,11 +1,13 @@
 package main
 
 // Set is Not exported
```
marcelocantos: Misleading comment. If it's not meant to be exported, call it `set` (or `intSet` for clarity and to reduce the likelihood of a name collision).

sysl2/util.go
```diff
@@ -1,11 +1,13 @@
...
-type Set map[int]bool
+type Set map[int]struct{}
+
+var empty struct{}
```
marcelocantos: Just use the literal `struct{}{}` instead of a variable. Instances of `struct{}` are free.

(Side note: `const` would be better than `var` here.)

sysl2/grammarParser.go
```diff
@@ -14,7 +15,6 @@ type parser struct {
 type builder struct {
     arr    []string
     tokens map[string]int32
```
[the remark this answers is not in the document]
sahejsingh: Atom.Id (from proto, generated code) is defined as int32

sysl2/util.go
sahejsingh: How do you declare a const struct{} ?

sysl2/sysl_parser.go
sahejsingh: its actually going to be renamed later to parseGrammar() as I plan to return ast as well.

sysl2/grammarParser.go
```diff
@@ -40,17 +40,16 @@ func (b *builder) handleChoice(choice *sysl.Choice) {
             if str != "" {
                 _, has := b.tokens[str]
                 if !has {
```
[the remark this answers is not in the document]
sahejsingh: really?. I like it this way as its clearer.

sysl2/util.go
sahejsingh: used struct{}{}

sysl2/util.go
sahejsingh: changed to intSet.

sysl2/sysl_parser.go
sahejsingh: renamed to parseGrammar()

sysl2/grammarParser.go
sahejsingh: Agreed on len() check. Moved it out of the loop.
The idea of `i + 1`  is to check if we can access the next element too. To me this seems more natural. But not a big deal.

sysl2/grammarParser.go
```diff
@@ -184,7 +182,7 @@ func buildFirstFollowSet(g *sysl.Grammar) (map[string]*Set, map[string]*Set) {
                 for i := range seq.Term {
                     // follow
                     l := len(seq.Term)
```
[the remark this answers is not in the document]
sahejsingh: moved it out of loop.

sysl2/Grammar1_test.go
```diff
@@ -33,29 +34,34 @@ func TestGrammar(t *testing.T) {
...
+
 func TestGrammar1(t *testing.T) {
     tokens := []int{2, 1, 1, 1, 0, -1}
     text := "bcccd"
     testParser(makeGrammar1(), 4, tokens, text, true, t)
 }
+
 func TestGrammarIncompleteInput(t *testing.T) {
     tokens := []int{2, 3, -1}
     text := "ba"
```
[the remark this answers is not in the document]
sahejsingh: added extra line to all functions.

juliaogris:
Good morning!
@sahejsingh what an accomplishment!
Really, you should be proud of what you have pulled off here, great starting point for Sysl2 and it didn't take you that long at all.

A few small points of order
* You could reopen this PR, edit it and change the target branch to `master` rather than creating a new PR
* CI on Appveyor seems broken in an unrelated way :(
* I've got a few minor comments on the code too but I'll wait until you have decided on whether you start a new PR or re-open this one.

sahejsingh:
This IS the 2nd pr and it is for master :)

sysl2/Grammar1_test.go
```diff
@@ -0,0 +1,253 @@
...
+    text := "bcccd"
+
+    testParser(makeGrammar1(), 4, tokens, text, true, t)
+}
+
+func TestGrammarIncompleteInput(t *testing.T) {
+    tokens := []int{2, 3, -1}
+    text := "ba"
+
+    testParser(makeGrammar1(), 4, tokens, text, false, t)
```
juliaogris: I'm a bit confused about what the `4` is. You call it `numTerms ` above, but it doesn't seem to change depending on the input text, so it seems to be something inherent in the grammar definition itself? If that's true that I think it would have been better to write a separate test that only checks for qualities of the grammar without input text.

sysl2/sysl_parser.go
```diff
@@ -0,0 +1,141 @@
...
+
+func getMinMaxCount(t *sysl.Term) (int, int) {
+    if t.Quantifier == nil {
+        return 1, 1
+    }
+
+    switch t.Quantifier.Union.(type) {
+    case *sysl.Quantifier_Optional: // 0 or 1
+        return 0, 1
+    case *sysl.Quantifier_ZeroPlus: // *
```
juliaogris: Why not call it `Quantifier_Star` ?

sysl2/Grammar1_test.go
```diff
@@ -0,0 +1,253 @@
...
+    tokens := []int{6, 0, 6, 2, 6, -1}
+
+    testParser(makeEXPR(), 7, tokens, text, true, t)
+}
+
+func TestOBJ(t *testing.T) {
+    text := "{}"
+    tokens := []int{0, 2, -1}
+
+    testParser(makeRepeatSeq(makeQuantifierZeroPlus()), 4, tokens, text, true, t)
```
juliaogris: I understand that `makeRepeatSeq ` creates a grammar for {} | { NUM (, NUM)*}, but I didn't manage to work out what `makeQuantifierZeroPlus` actually does.

PR #150, opened 2018-07-02
Remove unused sysl.util.jsonutil

fcostin:
Increases pytest coverage 67% -> 69%

- remove `sysl.util.jsonutil` . I believe this to be unused - can see no trace of the token "jsonutil" mentioned in any other version controlled file. It looks like this file was added in the first commit. Perhaps it used to be used but is used no longer, or secretly imported via some automagic plugin system. If it isn't used, deleting it is a quick win for increasing test coverage.

@anz-bank/sysl-developers


juliaogris:
@fcostin , @anz-rfc - sorry I updated the branch! You might want to merge this yourself (@anz-rfc should be a sysl developer and thus have rights to merge)

PR #158, opened 2018-11-04
Lang guide

sahejsingh:
Initial version of Sysl language guide.

@anz-bank/sysl-developers


docs/language_guide.md
```diff
@@ -0,0 +1,619 @@
...
+Sysl allows you to specify [Application](#application) behaviour and [Data Model](#data) that is shared between your applications. Another related concept is of software [Projects](#projects) where you can document what changes happened in each project or a release.
+
+More on this below.
+
+### Application
+An application is an independent entity that provides services via its various `endpoints`. E.g. a `MobileApp` allows users to `Login`, `Search` or `Order` items.
+
+Let's start to define a mobile app with these high level endpoints (or flows):
+```
+MobileApp:
```
anz-rfc: if `MobileApp` happens to be an application, could we make it clearer that this top-level thing is an application?

docs/language_guide.md
```diff
@@ -0,0 +1,619 @@
...
+  Login: ...
+  Search: ...
+  Order: ...
+```
+In the above example, `:` and `...` have special meaning. `:` is used to create a parent-child relationship.
+
+`MobileApp` is a user-defined name that `has` three `endpoints` viz `Login`, `Search` and `Orders`. The way to do that is by having a `newline` and `indent` after `:`.
```
anz-rfc: maybe `has` should be typeset as has

docs/language_guide.md
```diff
@@ -0,0 +1,619 @@
...
+
+The `...` (aka shortcut) means, we don't have enough details yet to describe how this endpoint behaves. Sysl allows you to take an iterative approach in documenting the behaviour. You add more as you know more.
```
anz-rfc: copy edit: "means, we" --> "means that we"

docs/language_guide.md
```diff
@@ -0,0 +1,619 @@
...
+### Data
+You will have various kinds of data passing through your systems. Sysl allows you to express ownership, information classification and other attributes of your data in one place.
+
+Continuing with the previous example, let's define a `Server` that expects `LoginData` for the `Login` Flow.
+```
+Server:
+  !type LoginData
+    username <: string
+    password <: string
+```
```
anz-rfc: not sure if this is the clearest example. in previous example we had a `Login` endpoint / flow. Is `LoginData` of this `Server` related to that, or a different `Login`?

docs/language_guide.md
```diff
@@ -0,0 +1,619 @@
...
+In the above example, `:` and `...` have special meaning. `:` is used to create a parent-child relationship.
+
+`MobileApp` is a user-defined name that `has` three `endpoints` viz `Login`, `Search` and `Orders`. The way to do that is by having a `newline` and `indent` after `:`.
+
+The `...` (aka shortcut) means, we don't have enough details yet to describe how this endpoint behaves. Sysl allows you to take an iterative approach in documenting the behaviour. You add more as you know more.
+
+### Data
+You will have various kinds of data passing through your systems. Sysl allows you to express ownership, information classification and other attributes of your data in one place.
+
+Continuing with the previous example, let's define a `Server` that expects `LoginData` for the `Login` Flow.
```
anz-rfc: what is `Server`? is it a distinct application which happens to have the name `Server`?

docs/language_guide.md
```diff
@@ -0,0 +1,619 @@
...
+  *  CustomerOrder has a foreign key customer_id which refers to the primary key of Customer (i.e. customer_id).
+
+See [data.sysl](assets/data.sysl) for complete example for Relation model.
+
+![](assets/data-Relational-Model.png)
+
+#### Object Model
+Define a typical in memory Object model of an application
+```java
+CustomerOrderModel:
```
anz-rfc: i'm still confused about the syntax. is this top level `CustomerOrderModel` also an "application"?

docs/language_guide.md
```diff
@@ -0,0 +1,619 @@
...
+```
+
+and you use `import` in `client.sysl`
+
+```js
+import server
+
+Client:
+  Login:
+    Server <- Login
```
anz-rfc: sysl-language complaint -- i'm offended that symbols get imported into the global namespace.

Now I have to exhaustively read all files that are (transitively?) imported to find where `Server` is defined. or search across the project for `Server:`, i guess. or perhaps use some smart sysl language IDE plugin which can take me to the definition.

docs/language_guide.md
```diff
@@ -0,0 +1,619 @@
...
+Sysl analyzes the starting endpoint and finds all the `call`s that this endpoint makes to other endpoints (including the ones to other applications). It finds all the transitive dependencies till there are none.
+
+In the above diagram, `DB` is the last app in this flow. Sysl also captures the return data that each endpoint returns to its caller. See below for more details.
+
+#### Return response
+Express returning response to the caller by using the `return` keyword. See the example above on how to use the `return` keyword.
+```
+App:
+  Endpoint:
+    return STRING | TYPE
```
anz-rfc: i assume this isn't an example of valid syntax ( STRING | TYPE). it might be clearer to give an example of a string, and another example involving a type. or separately specify grammar so it doesnt look like an example

docs/language_guide.md
```diff
@@ -0,0 +1,619 @@
...
+```
+In the above example, everything after `return` keyword till the end-of-line is considered response. You can have:
+  * STRING - a descrition of what is returned
```
anz-rfc: spellcheck  descrition --> description

might be a good idea to spellcheck the entire file

docs/language_guide.md
```diff
@@ -0,0 +1,619 @@
...
+You can express an endpoint's critical decisions using IF/ELSE statement:
+
+```
+Server:
+  HandleFormSubmit:
+    validate input
+    IF session exists:
+      use existing session
+    Else:
+      create new session
```
anz-rfc: IF / Else

maybe guide could state if sysl is case sensitive or not

docs/language_guide.md
```diff
@@ -0,0 +1,619 @@
+# Language Guide
```
anz-rfc: perhaps it'd be good to have both tutorial style documentation (like this) and language reference documentation.

i'd probably be able to understand more of what was going on here if i was able to read this at the same time as reading some language reference

anz-rfc:
@anzdaddy asked me to look at this again

thanks @sahejsingh for making some changes in response to my feedback earlier . it doesnt seem like much  has changed since i reviewed it earlier. is it worth talking through my other feedback (e.g. lack of explanation about what top-level elements in the sysl files are) that hasn't been addressed by the most recent commit?

sahejsingh:
> dumb questions: is there anything special about the word "MobileApp"? how does sysl know "MobileApp" is an application?
> 
> maybe could say something like "the name MobileApp is user-defined and has no special meaning to sysl" or perhaps, "in sysl, any top-level word followed by a colon such as "MobileApp:" or "Foo:" or "Banana:" defines a an application" or perhaps "a named application".
> 
> that probably isnt clearest way to explain it, hopefully you roughly get what i'm getting at

@anz-rfc , the explanation clearly states the following:
> To explain these concepts, we will design an application called MobileApp which interacts with another application called Server.

and later...
> MobileApp is a user-defined Application that does not have any endpoints yet. We will design this app as we move along.

I think it is clear that user is defining an app called 'MobileApp'.

I have also shows that a name like 'MobileApp' followed by ':' and an indent makes it an application.

PR #168, opened 2019-01-06
:art: Pass abstract filesystem to sysl.Parse() + more testable main()

anzdaddy:
Changes proposed in this pull request:
- Replace direct os calls in `sysl.Parse` with `http.FileSystem` parameter.
- Split main into three functions to improve support for testing.

@anz-bank/sysl-developers

PR #169, opened 2019-01-06
🎨 Cleaner parser tests

anzdaddy:
- Use computed temp filenames instead of g{olden,enerated}.txt.
- Enable golden-output regression tests.

Builds on yet-to-be-merged #168. Only commits after 979fc8c are specific to this PR.

@anz-bank/sysl-developers

### DIFF
--- a/sysl2/sysl/Parser_test.go
+++ b/sysl2/sysl/Parser_test.go
@@ -131,14 +131,19 @@ func parseAndCompareWithPython(filename, root string) (bool, error) {
 		return false, err
 	}
 	defer golden.Close()
-	defer os.Remove(golden.Name())
 
 	pyModule, err := pyParse(filename, root, golden.Name())
 	if err != nil {
+		fmt.Printf("%#v retained for checking", golden.Name())
 		return false, err
 	}
 
-	return parseAndCompare(filename, root, golden.Name(), pyModule)
+	equal, err := parseAndCompare(filename, root, golden.Name(), pyModule)
+	if err != nil {
+		fmt.Printf("%#v retained for checking", golden.Name())
+	}
+	os.Remove(golden.Name())
+	return equal, err
 }
 
 func parseAndCompareWithGolden(filename, root string) (bool, error) {
--- a/sysl2/sysl/Parser_test.go
+++ b/sysl2/sysl/Parser_test.go
@@ -31,12 +31,12 @@ func readSyslModule(filename string) (*sysl.Module, error) {
 	return module, nil
 }
 
-func pySysl() string {
-	if pySysl, found := os.LookupEnv("SYSL_PYTHON_BIN"); found {
+var pySysl = func() string {
+	if pySysl, ok := os.LookupEnv("SYSL_PYTHON_BIN"); ok {
 		return pySysl
 	}
 	return "sysl"
-}
+}()
 
 func pyParse(filename, root, output string) (*sysl.Module, error) {
 	args := []string{"textpb", "-o", output, filename}
@@ -46,11 +46,11 @@ func pyParse(filename, root, output string) (*sysl.Module, error) {
 		args = append(rootArg, args...)
 	}
 
-	cmd := exec.Command(pySysl(), args...)
+	cmd := exec.Command(pySysl, args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	if err := cmd.Run(); err != nil {
-		return nil, fmt.Errorf("Error running %#v %#v: %s", pySysl(), args, err)
+		return nil, fmt.Errorf("Error running %#v %#v: %s", pySysl, args, err)
 	}
 
 	return readSyslModule(output)
@@ -124,7 +124,7 @@ func parseAndCompare(filename, root, golden string, goldenModule *sysl.Module) (
 }
 
 func parseAndCompareWithPython(filename, root string) (bool, error) {
-	fmt.Printf("%35s <=> $(%s %[1]s)\n", filename, pySysl())
+	fmt.Printf("%35s <=> $(%s %[1]s)\n", filename, pySysl)
 
 	golden, err := ioutil.TempFile("", "sysl-test-golden-*.textpb")
 	if err != nil {
--- a/sysl2/sysl/Parser_test.go
+++ b/sysl2/sysl/Parser_test.go
@@ -42,7 +42,6 @@ func pyParse(filename, root, output string) (*sysl.Module, error) {
 	args := []string{"textpb", "-o", output, filename}
 	if len(root) > 0 {
 		rootArg := []string{"--root", root}
-		output = args[2]
 		args = append(rootArg, args...)
 	}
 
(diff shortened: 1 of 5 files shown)


sysl2/sysl/Parser_test.go
```diff
@@ -70,36 +78,71 @@ func parseComparable(filename, root string) (*sysl.Module, error) {
...
+	}
+
+	generated, err := ioutil.TempFile("", "sysl-test-*.textpb")
 	if err != nil {
 		return false, err
 	}
-	if proto.Equal(pyModule, goModule) {
-		return true, nil
-	}
+	defer generated.Close()
```
anzboi: `generated.Close()` is called further down on line 105. The deferred call is generating an error but is ignored due to defer.

sysl2/sysl/Parser_test.go
```diff
@@ -111,171 +154,181 @@ func parseAndPrint(t *testing.T, filename, root string) error {
 	return nil
 }
 
-func testParse(t *testing.T, filename, root string) {
-	equal, err := parseAndCompare(filename, root)
+func testParseAgainstPython(t *testing.T, filename, root string) {
```
anzboi: root is only non-empty in two of the tests below (lines 263 and 329). In all other cases, the filename is given with the path. Is root something we want to keep for future or do we get want to get rid of it?

sysl2/sysl/Parser_test.go
```diff
@@ -111,171 +154,181 @@ func parseAndPrint(t *testing.T, filename, root string) error {
...
+	equal, err := parseAndCompareWithPython(filename, root)
+	assert.NoError(t, err)
+	assert.True(t, equal, "Mismatch")
+}
+
+func testParseAgainstGolden(t *testing.T, filename, root string) {
```
anzboi: Similar here, except root is never non-empty.

sysl2/sysl/Parser_test.go
sahejsingh: root is required for import statements to work.

sysl2/sysl/Parser_test.go
```diff
@@ -13,39 +14,46 @@ import (
...
+
+	module := &sysl.Module{}
+	if err := proto.UnmarshalText(buf.String(), module); err != nil {
+		return nil, err
+	}
+	return module, nil
+}
+
+func pySysl() string {
+	if pySysl, found := os.LookupEnv("SYSL_PYTHON_BIN"); found {
```
juliaogris: Nit: I believe go convention is to call the var above `ok` not `found` (see [docs](https://golang.org/pkg/os/#LookupEnv))

sysl2/sysl/Parser_test.go
```diff
@@ -13,39 +14,46 @@ import (
...
+	return "sysl"
+}
+
+func pyParse(filename, root, output string) (*sysl.Module, error) {
+	args := []string{"textpb", "-o", output, filename}
 	if len(root) > 0 {
 		rootArg := []string{"--root", root}
-		// TODO: This looks dubious
-		args[2] = root + "/" + args[2]
 		output = args[2]
```
juliaogris: Remove?

sysl2/sysl/Parser_test.go
```diff
@@ -70,36 +78,71 @@ func parseComparable(filename, root string) (*sysl.Module, error) {
 	return module, nil
 }
 
-func parseAndCompare(filename, root string) (bool, error) {
-	goModule, err := parseComparable(filename, root)
+func parseAndCompare(filename, root, golden string, goldenModule *sysl.Module) (bool, error) {
+	module, err := parseComparable(filename, root)
```
juliaogris: Not part of this PR, but seeing as it's a cleanup: remove line 69 above (`// app.SourceContext.Start.Col = 0`)

sysl2/sysl/Parser_test.go
```diff
@@ -70,36 +78,71 @@ func parseComparable(filename, root string) (*sysl.Module, error) {
...
 
-	golden := "golden.txt"
-	generated := "generated.txt"
-	if err = TextPB(pyModule, golden); err != nil {
+	if err = TextPB(goldenModule, golden); err != nil {
 		return false, err
 	}
-	if err = TextPB(goModule, generated); err != nil {
+
+	if err = FTextPB(generated, module); err != nil {
```
juliaogris: `generated` (tempfile) first used here, define right above (I think)

sysl2/sysl/Parser_test.go
```diff
@@ -70,36 +78,71 @@ func parseComparable(filename, root string) (*sysl.Module, error) {
...
+
+	generated, err := ioutil.TempFile("", "sysl-test-*.textpb")
 	if err != nil {
 		return false, err
 	}
-	if proto.Equal(pyModule, goModule) {
-		return true, nil
-	}
+	defer generated.Close()
+	defer os.Remove(generated.Name())
```
juliaogris: Are you sure you want to remove the tempfile even if the tests fails? Might be worth keeping it around for investigation.

sysl2/sysl/Parser_test.go
anzdaddy: Both calls are needed and it's easier to call `Close` redundantly than to check whether it has already been called. I could add a check or simply an explanatory comment.

sahejsingh: uh. more complicated code with defer. reminds me of c++ destructors/RAII. Goes to show error handling is not that simple. You end up having multiple code paths to think about.

@marcelocantos Is the intent of `defer close` to close the file in case of returning from  99 and 103 and the close at 103 is to make sure its closed before diff?

anzdaddy: Done (added a check).

sysl2/sysl/Parser_test.go
anzdaddy: Good idea! Done.

sysl2/sysl/Parser_test.go
anzdaddy: Done.

sysl2/sysl/Parser_test.go
anzdaddy: As above.

sysl2/sysl/Parser_test.go
anzdaddy: I've also done it in `parseAndCompareWithPython`.

sysl2/sysl/Parser_test.go
anzdaddy: Done. I also reworked it into a computed-once variable instead of a function call.

sysl2/sysl/Parser_test.go
juliaogris: Sorry, I must be having a completely dumb moment, but isn't `output` already equal to `args[2]`

anzdaddy: Ah yes, I see now. Off-by-one error in my brain.

PR #173, opened 2019-01-06
✨ Introduce "sequence of" types

anzdaddy:
These are basically array types. I've avoided introducing these to encourage using sets instead, but they are needed in some scenarios.

@anz-bank/sysl-developers


sysl2/sysl/grammar/SyslParser.g4
```diff
@@ -23,8 +23,10 @@ user_defined_type       : name_str;
 types: user_defined_type | reference | NativeDataTypes;
 set_of: SET_OF | (E_SET_OF);
 set_type            : set_of types size_spec?;
+sequence_of: SEQUENCE_OF | (E_SEQUENCE_OF);
+sequence_type            : sequence_of types size_spec?;
 //TODO : allow for other collection types?
```
sahejsingh: got one more collection type :)
Yay!

PR #177, opened 2019-01-16
:bug: Call Getenv("SYSL_LEXER_LOG") once per run, not once per call

marcelocantos:
@anz-bank/sysl-developers

### DIFF
--- a/sysl2/sysl/grammar/lexer_impl.go
+++ b/sysl2/sysl/grammar/lexer_impl.go
@@ -8,6 +8,8 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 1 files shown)


sysl2/sysl/grammar/lexer_impl.go
```diff
@@ -116,7 +118,7 @@ func GetNextToken(l *SyslLexer) antlr.Token {
 	}
 
 	next := l.BaseLexer.NextToken()
-	if os.Getenv("SYSL_LEXER_LOG") != "" {
+	if syslLexerLog {
```
anz-rfc: lol, it was getting the env var once every time something called `GetNextToken` was called. i bet that was a lot of lookups...

PR #193, opened 2019-02-19
Feature/go-tags

kasunfdo:
- Implemented null comparison
- Generate Go tags, capitalizing common initialisms


sysl2/sysl/examples/grammars/go/go.gen.g
```diff
@@ -5,7 +5,7 @@ ImportDecl: 'import' '(\n' ImportSpec* '\n)\n';
 ImportSpec: Import '\n';
 TopLevelDecl: Declaration | FunctionDecl | MethodDecl;
 Declaration: VarDecl | ConstDecl | StructType | InterfaceType ;
-StructType : 'type' StructName 'struct' '{\n' FieldDecl* '}\n';
+StructType : Comment? '\n' 'type' StructName 'struct' '{\n' FieldDecl* '}\n';
```
anz-rfc: Comments can probably go in many places -- it seems a bit strange to just have comment as an optional prefix of Struct in the grammar.


for example, it looks like Go's AST machinery has a map from AST nodes to comments:

https://golang.org/pkg/go/ast/#NewCommentMap

sysl2/sysl/examples/grammars/go/go.gen.sysl
```diff
@@ -18,6 +18,12 @@ CodeGenTransform:
         else t.type_ref
     )
 
+  !view GetGoField(word <: string) -> string:
+    word -> (:
+      let wordUpper = ToTitle(word)
+      out = if (wordUpper in {"ACL", "API", "ASCII", "CPU", "CSS", "DNS", "EOF", "GUID", "HTML", "HTTP", "HTTPS", "ID", "IP", "JSON", "LHS", "QPS", "RAM", "RHS", "RPC", "SLA", "SMTP", "SQL", "SSH", "TCP", "TLS", "TTL", "UDP", "UI", "UID", "UUID", "URI", "URL", "UTF8", "VM", "XML", "XMPP", "XSRF", "XSS"}) == true then wordUpper else Title(ToLower(word))
```
anz-rfc: feature request: ability for whoever is running codegen to override rules for field naming.

suggestion:   instead of this heuristic could we add support for optionally specifying the go field names in the input sysl file, and using them.

why? each specific application domain or business domain may have its own collection of commonly used acronyms. so its not possible to have one list of acryonyms that does the right thing for all users.  for example, in telco domain, some telco companies use FAT as an acronym for a bit of network equipment, whereas in many other contexts "fat" would be a normal word.

sysl2/sysl/examples/grammars/go/go.gen.sysl
```diff
@@ -26,16 +32,19 @@ CodeGenTransform:
...
     fields -> (field:
       let isKeyword = IsKeyword(field.key).out
-      let prefix = if isKeyword == true then "_" else ""
-      identifier = prefix + field.key
+      let fieldName = GetGoField(field.key).out
+      identifier = fieldName
       let IsPtr = if field.value.optional == true then "*" else ""
       Type = IsPtr + GoType(field.value).out
+      let fieldTag = '`json:"' + ToLower(fieldName)
+      Tag = if field.value.optional == true then (fieldTag + ',omitempty"`') else (fieldTag + '"`')
```
anz-rfc: feature request ---- probably beyond scope of your change here:

it would be good to allow user who is running codegen to add or customise field tagging for new kinds of tags. these might be user defined tags with some user defined meaning specific to the application.

for example, in addition to `json:` field tags there is another project that uses field tags to declare validations -- https://github.com/go-validator/validator#usage

more tags here: https://github.com/golang/go/wiki/Well-known-struct-tags

sysl2/sysl/examples/grammars/go/go.gen.sysl
```diff
@@ -26,16 +32,19 @@ CodeGenTransform:
...
     )
 
   !view GoStructs(types <: set of Type) -> set of TopLevelDecl:
     types -> (type:
         Declaration = type -> <Declaration>(:
           StructType = type -> <StructType>(:
+            Comment = "\n//" + .key + " ..."
```
anz-rfc: in current form it would be better not to have comment, if comment is just "<StructName> ...".

sysl2/sysl/exprOp.go
```diff
@@ -139,6 +150,10 @@ func evalBinExpr(op sysl.Expr_BinExpr_Op, lhs, rhs *sysl.Value) *sysl.Value {
 func addBinaryOps() {
 	var ops = []OpArgs{
 		OpArgs{sysl.Expr_BinExpr_Op_name[int32(sysl.Expr_BinExpr_EQ)], VALUE_BOOL, VALUE_BOOL, cmpBool},
+		OpArgs{sysl.Expr_BinExpr_Op_name[int32(sysl.Expr_BinExpr_EQ)], VALUE_NULL, VALUE_NULL, cmpNullTrue},
+		OpArgs{sysl.Expr_BinExpr_Op_name[int32(sysl.Expr_BinExpr_NE)], VALUE_NULL, VALUE_NULL, cmpNullFalse},
+		OpArgs{sysl.Expr_BinExpr_Op_name[int32(sysl.Expr_BinExpr_EQ)], VALUE_STRING, VALUE_NULL, cmpNullFalse},
+		OpArgs{sysl.Expr_BinExpr_Op_name[int32(sysl.Expr_BinExpr_NE)], VALUE_STRING, VALUE_NULL, cmpNullTrue},
```
anz-rfc: ok, makes sense.

sysl2/sysl/examples/grammars/go/go.gen.sysl
```diff
@@ -26,16 +32,19 @@ CodeGenTransform:
   !view GoStructFields(fields <: set of Type) -> set of FieldDecl:
     fields -> (field:
       let isKeyword = IsKeyword(field.key).out
-      let prefix = if isKeyword == true then "_" else ""
-      identifier = prefix + field.key
+      let fieldName = GetGoField(field.key).out
+      identifier = fieldName
```
anz-rfc: might be a bug:   do we still need to do `isKeyword` check here on our computed identifier, and rename identifier if it collides with a keyword?  or if collision is no longer possible, why are we calling `IsKeyword`?

sysl2/sysl/examples/grammars/go/go.gen.sysl
marcelocantos: It's to shut the linter up.

sysl2/sysl/examples/grammars/go/go.gen.sysl
kasunfdo: Hopefully identifiers will not collide with keywords as we capitalize first letter of identifiers. We are not calling keyword anymore

sysl2/sysl/examples/grammars/go/go.gen.g
kasunfdo: This is used just to avoid inter warnings.

sysl2/sysl/examples/grammars/go/go.gen.sysl
kasunfdo: Thought of that as well. Will add them in next phase

sysl2/sysl/examples/grammars/go/go.gen.sysl
kasunfdo: Yes I agree. For the moment this is added to avoid linter complains.

sysl2/sysl/examples/grammars/go/go.gen.g
kasunfdo: Yes I agree. For the moment this is added to avoid linter complains.

sysl2/sysl/examples/grammars/go/go.gen.sysl
kasunfdo: Yes, in transform, you can add anything and it will appear on generated code. This is just an example of generating json tags.

sysl2/sysl/examples/grammars/go/go.gen.sysl
kasunfdo: Users can define their own list of naming conventions in transform and the code will be generated as expected.

sysl2/sysl/examples/grammars/go/go.gen.g
sahejsingh: Comments are generally not part of the grammar. Have a look here: https://golang.org/ref/spec#Struct_types
We are doing the reverse of what generally grammars are used for ie. generating text instead of parsing them. So don't be surprised with these new constructs.

The point of this grammar is to assist you in producing the text in the structure that you want.

If you want to have multiple comments, for eg. struct comment, field comment, they will have their own place in the grammar.

sysl2/sysl/examples/grammars/go/go.gen.sysl
sahejsingh: the whole purpose of the grammar and the transform files is to be able to write your own rules.

anz-rfc: okay, thanks for explaining.

sysl2/sysl/examples/grammars/go/go.gen.sysl
anz-rfc: okay, thanks for explaining.

anzdaddy: #203

PR #199, opened 2019-02-25
Encode Rest API's URL and Query Params separately

sahejsingh:
Encode Rest API's URL and Query params separately.
Updated grammar: removed query_params from http_part. Query params are only supported in the method definition.
@anz-bank/sysl-developers


src/proto/sysl.proto
```diff
@@ -104,15 +105,7 @@ message Endpoint {
...
-            bool    loc  = 4;
-
-            enum Loc {
-                NO_In = 0;
-                QUERY = 1;
-                PATH = 2;
-                BODY = 3;
-            }
-
+            bool    loc  = 4; // no longer used
```
anzdaddy: Needs to be renamed to `DEPRECATED_loc`. Leave it for #200.

PR #206, opened 2019-03-08
Where map

sahejsingh:
Add whereMap to filter on map elements.

@anz-bank/sysl-developers


sysl2/naive/grammarParser.go
```diff
@@ -111,7 +111,7 @@ func GetTermMinMaxCount(t *sysl.Term) (int, int) {
 
 func (p *parser) parse(g *sysl.Grammar, input int, val interface{}) (bool, int, []interface{}) {
 	if input == len(*(p.tokens)) {
-		logrus.Debugln("got input length zero!!!")
+		logrus.Debug("got input length zero!!!")
```
anzdaddy: #207

sysl2/sysl/exprOp.go
```diff
@@ -33,6 +33,10 @@ func neInt64(lhs, rhs *sysl.Value) *sysl.Value {
 	return MakeValueBool(lhs.GetI() != rhs.GetI())
 }
 
+func neString(lhs, rhs *sysl.Value) *sysl.Value {
+	return MakeValueBool(lhs.GetS() != rhs.GetS())
+}
```
anzdaddy: Suggestion (easier to show than explain):
```go
func not(f func(lhs, rhs *sysl.Value) *sysl.Value) func(lhs, rhs *sysl.Value) *sysl.Value {
	return func(lhs, rhs *sysl.Value) *sysl.Value {
		return MakeValueBool(!(f(lhs, rhs)))
	}
}
```
With this in place, use `not(cmpString)` in lieu of writing and using `neString`, and likewise with other negations (`not(cmpInt)` vs `neInt64`, `not(gtInt64)` vs `leInt64`, …).

sysl2/sysl/binexprEval.go
```diff
@@ -113,9 +113,16 @@ func (op Eval_LHS_Over_RhsStrategy) eval(txApp *sysl.Application, assign *Scope,
 	vType := getValueType(lhs_v)
 	itemType := getContainedType(lhs_v)
 	key := makeKey(binexpr.Op, vType, itemType)
-
+	scopeVarValue, hasScopeVar := (*assign)[binexpr.Scopevar]
 	if f, has := exprFunctions[key]; has {
-		return f(txApp, assign, lhs_v, binexpr.Scopevar, binexpr.Rhs)
+		var result *sysl.Value
+
+		result = f(txApp, assign, lhs_v, binexpr.Scopevar, binexpr.Rhs)
```
anzdaddy:
```go
result := …
```

sysl2/sysl/binexprEval.go
```diff
@@ -113,9 +113,16 @@ func (op Eval_LHS_Over_RhsStrategy) eval(txApp *sysl.Application, assign *Scope,
 	vType := getValueType(lhs_v)
 	itemType := getContainedType(lhs_v)
 	key := makeKey(binexpr.Op, vType, itemType)
-
+	scopeVarValue, hasScopeVar := (*assign)[binexpr.Scopevar]
```
anzdaddy: `assign` doesn't need to be a `*Scope`. It can be a `Scope`. Maps are already reference types.

PR #212, opened 2019-03-27
Modify travis config and publish sysl2 image

kasunfdo:
Changes:

- Modify repo name in travis configs (Fixes travis skipping deployment)
- Build image with sysl2 on travis

### DIFF
--- a/Dockerfile-sysl2-ubuntu
+++ b/Dockerfile-sysl2-ubuntu
@@ -10,3 +10,4 @@ ADD dist/sysl-*.whl /sysl
 ADD gosysl/gosysl-linux /usr/bin/sysl2
 
 RUN pip install sysl-*.whl
+RUN ln -s /usr/bin/sysl2 /usr/bin/syslgen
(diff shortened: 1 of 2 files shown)


Dockerfile-sysl2-ubuntu
```diff
@@ -0,0 +1,13 @@
...
+  apt-get -qq upgrade -y --fix-missing && \
+  apt-get -qq install -y python2.7 python-pip
+  
+WORKDIR /sysl
+
+ADD dist/sysl-*.whl /sysl
+ADD gosysl/gosysl-linux /sysl
+
+RUN pip install sysl-*.whl
+RUN ln -s /sysl/gosysl-linux /usr/bin/syslgen
```
sahejsingh: we could rename the `gosysl-*` binaries to `sysl2`.
And then add `sysl2` in the path

sahejsingh: why not just copy them to the `/usr/bin`?

kasunfdo: Agreed. Renamed as sysl2 and copied to /usr/bin

PR #222, opened 2019-04-04
Modify swagger importer with few fixes

kasunfdo:
- Fixed return value and body param
- Add a unit test to cover changes


kasunfdo:
Require more fixes according to bank-anywhere spec

Issue #233, opened 2019-06-12
Go SYSL SD cmd: remove -modules arg 


sahejsingh:
Following command line args are incompatible with python sysl sd.

- -endpoint instead of -s
- -module instead of the last arg in the commandline
- -o does not work. Have to specify -output.
- --appfmt

### Expected behaviour
The command line options are not compatible with the python version of sysl.
Following command should work:
`sd -s 'MobileApp <- Login' -output call.png  call.sysl`

jecofang01:
Seems like this issue can be closed as it already be fixed in the PR #234

PR #234, opened 2019-06-12
Fix issue #232 and #233

### LINKED ISSUE #232: go sysl sd produces empty png file.
cd to  repo, `sysl/docs/assets` directory:

Python sysl sd command produces the correct output:
`sysl sd -s 'MobileApp <- Login' -o call.png call`

But the equivalent go-sysl-sd command:

`sd -endpoint 'MobileApp <- Login' -output call.png -modules call.sysl`

produces an empty file `call.png`.

### Expected behaviour
The above mentioned command should generate the correct sequence diagram

Another issue:
The python and go versions do not have the same command line options.

loki5265:
* Add default value for SYSL_PLANTUML
* Add kingpin package to handle command-line args
* Align command-line args with python

Fixes #232, #233 .

-
-

@anz-bank/sysl-developers

### DIFF
--- a/sysl2/sysl/seqs.go
+++ b/sysl2/sysl/seqs.go
@@ -10,6 +10,7 @@ import (
 	"github.com/anz-bank/sysl/src/proto"
 	"github.com/anz-bank/sysl/sysl2/sysl/seqs"
 	log "github.com/sirupsen/logrus"
+	"gopkg.in/alecthomas/kingpin.v2"
 )
 
 type sequenceDiagParam struct {
@@ -27,19 +28,6 @@ func generateSequenceDiag(m *sysl.Module, p *sequenceDiagParam) (string, error) 
 	e.Accept(v)
 
 	return w.String(), nil
-}
-
-type arrayFlags []string
-
-// String implements flag.Value.
-func (i *arrayFlags) String() string {
-	return strings.Join(*i, ",")
-}
-
-// Set implements flag.Value.
-func (i *arrayFlags) Set(value string) error {
-	*i = append(*i, value)
-	return nil
 }
 
 func loadApp(root string, models string) *sysl.Module {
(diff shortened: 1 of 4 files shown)


sysl2/sysl/seqs.go
```diff
@@ -147,34 +135,36 @@ func DoGenerateSequenceDiagrams(stdout, stderr io.Writer, flags *flag.FlagSet, a
...
-	endpoint_format := flags.String("endpoint_format", "%(epname)", "Specify the format string for sequence diagram endpoints. "+
-		"May include %%(epname), %%(eplongname) and %%(@foo) for attribute foo(default: %(epname))")
-	app_format := flags.String("app_format", "%(appname)", "Specify the format string for sequence diagram participants. "+
-		"May include %%(appname) and %%(@foo) for attribute foo(default: %(appname))")
-	title := flags.String("title", "", "diagram title")
-	plantuml := flags.String("plantuml", "", strings.Join([]string{"base url of plantuml server",
+	sd := kingpin.New("sd", "Generate sequence diagram")
+	root := sd.Flag("root", "sysl root directory for input model file (default: .)").Default(".").String()
+	endpoint_format := sd.Flag("endpoint_format", "Specify the format string for sequence diagram endpoints. "+
+		"May include %%(epname), %%(eplongname) and %%(@foo) for attribute foo(default: %(epname))").Default("%(epname)").String()
```
sahejsingh: %%(epname) is incorrect.
Only need 1 `%` character.

sysl2/sysl/seqs.go
```diff
@@ -147,34 +135,36 @@ func DoGenerateSequenceDiagrams(stdout, stderr io.Writer, flags *flag.FlagSet, a
...
+	app_format := sd.Flag("app_format", "Specify the format string for sequence diagram participants. "+
+		"May include %%(appname) and %%(@foo) for attribute foo(default: %(appname))").Default("%(appname)").String()
+	title := sd.Flag("title", "diagram title").Short('t').String()
+	plantuml := sd.Flag("plantuml", strings.Join([]string{"base url of plantuml server",
```
sahejsingh: is it possible to set the default value of plantuml here?

sysl2/sysl/seqs/diagutil.go
```diff
@@ -20,6 +20,10 @@ func OutputPlantuml(output, plantuml, umlInput string) {
 		plantuml = os.Getenv("SYSL_PLANTUML")
 	}
 
+	if plantuml == "" {
```
sahejsingh: Remove this and set default where you add the plantuml command line arg.

sysl2/sysl/seqs.go
```diff
@@ -147,34 +135,36 @@ func DoGenerateSequenceDiagrams(stdout, stderr io.Writer, flags *flag.FlagSet, a
...
+		"comma-list of shell globs) to limit the diagrams generated").Short('a').Strings()
+	blackboxes_flag := sd.Flag("blackbox", "Apps to be treated as black boxes").Strings()
 
 	// Following variables currently are unused. Keep them to align with the python version.
-	filter := flags.String("filter", "", "Only generate diagrams whose output paths match a pattern")
-	no_activations := flags.Bool("no-activations", true, "Suppress sequence diagram activation bars(default: true)")
-	verbose := flags.Bool("verbose", false, "Report each output(default: false)")
-	expire_cache := flags.Bool("expire-cache", false, "Expire cache entries to force checking against real destination(default: false)")
-	dry_run := flags.Bool("dry-run", false, "Don't perform confluence uploads, but show what would have happened(default: false)")
+	filter := sd.Flag("filter", "Only generate diagrams whose output paths match a pattern").String()
```
sahejsingh: remove these if we are not handling them.

sysl2/sysl/seqs.go
```diff
@@ -184,7 +174,7 @@ func DoGenerateSequenceDiagrams(stdout, stderr io.Writer, flags *flag.FlagSet, a
 	log.Debugf("no_activations: %t\n", *no_activations)
 	log.Debugf("endpoint_format: %s\n", *endpoint_format)
 	log.Debugf("app_format: %s\n", *app_format)
-	log.Debugf("blackbox: %s\n", blackboxes_flag)
+	log.Debugf("blackbox: %s\n", *blackboxes_flag)
```
sahejsingh: how do i run at debug level? there should be an option to change the log level.

sysl2/sysl/seqs.go
loki5265: Firstly I copy this description from python and I thought it's correct.
Delete 1 %.

sysl2/sysl/seqs.go
loki5265: It's possible. But I get another question. After setting default value here, I believe part of the code in diagutil.go should be changed too.
if plantuml == "" {
	plantuml = os.Getenv("SYSL_PLANTUML")
} If better to remove this, too?

sysl2/sysl/seqs/diagutil.go
loki5265: yep, remove it.

sysl2/sysl/seqs.go
loki5265: removed.

sysl2/sysl/seqs.go
loki5265: add "log" command line arg to handle this.

test-gosysl.sh
```diff
@@ -8,3 +8,10 @@ for f in $ROOT/*.sysl; do
 done;
 
 rm $ROOT/*.out.txt
+
+ln -s $GOPATH/bin/sysl $GOPATH/bin/sd
+$GOPATH/bin/sd -a 'Project' $ROOT/sequence_diagram_project.sysl
+rm _.png
```
sahejsingh: How do you know if the file is getting generated here?
there is no server running on localhost:8080.
and the code here https://github.com/anz-bank/sysl/blob/master/sysl2/sysl/seqs/diagutil.go#L19 does not use value of SYSL_PLANTUML because plantuml has a default value. right?

loki5265: I revert the logic to check the SYSL_PLANTUML. If no value from command arg, will use the value from SYSL_PLANTUML. If env not setted, will use default value of "http://localhost:8080/plantuml".
And actually I'm not sure how to run the binary test. I add a new arg value for plantuml with "http://www.plantuml.com/plantuml". Hope it work.

sysl2/sysl/seqs/diagutil.go
sahejsingh: all the checks on `plantuml` should move to where the command line args are evaluated. Not here.
One place where it could be is to move lines 19-24 to seqs.go

test-gosysl.sh
loki5265: yep. I've removed the -p param from the go-testsysl.sh file.

sysl2/sysl/seqs.go
```diff
@@ -147,55 +136,59 @@ func DoGenerateSequenceDiagrams(stdout, stderr io.Writer, flags *flag.FlagSet, a
...
+		log.SetLevel(level)
+	}
+
+	if *plantuml == "" {
+		*plantuml = os.Getenv("SYSL_PLANTUML")
+	}
+	if *plantuml == "" {
+		*plantuml = "http://localhost:8080/plantuml"
+	}
+
```
sahejsingh: small nit pick:
if you pass plantuml in the cmd line
then you are checking the value of plantuml twice.

sahejsingh:
```
if *plantuml == "" {
        if *plantuml, has = os.LookupEnv("SYSL_PLANTUML"); !has {
            *plantuml = "http://localhost:8080/plantuml"
        }
    }
```

PR #240, opened 2019-06-24
make swagger exporter callable from command line

sahejsingh:
Initial support to call swagger exporter.

@anz-bank/sysl-developers

### DIFF
--- a/src/sysl/importers/import_swagger.py
+++ b/src/sysl/importers/import_swagger.py
@@ -17,7 +17,7 @@ from sysl.util import writer
 from sysl.proto import sysl_pb2
 
 
-SWAGGER_FORMATS = {'int32', 'int64', 'float', 'double', 'date', 'date-time'}
+SWAGGER_FORMATS = {'int32', 'int64', 'float', 'double', 'date', 'date-time', 'integer'}
 
 SYSL_TYPES = {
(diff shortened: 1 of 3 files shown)


src/sysl/importers/import_swagger.py
```diff
@@ -213,7 +213,7 @@ def writeEndpoints(self, swag, w):
                         method.upper(),
                         ' ?' if params['query'] else '',
                         '&'.join(
-                            '{}={}{}'.format(p['name'], TYPE_MAP[p['type']], '' if p['required'] else '?')
+                            '{}={}{}'.format(p['name'], 'int' if p['type'] == 'number' else TYPE_MAP[p['type']], '' if p['required'] else '?')
```
anzdaddy: Not sure I'm reading this right. 'number' isn't always (or even usually) 'int'. If it's correct, it needs a comment.

sahejsingh: the problem here is the original code is only looking at the `type`. It should also look at `format` to decide sysl type of the query parameter.

I guess I have decided that for query params a `number` is always an integer.

I should remove this and introduce a bit more logic to decide on query param's type.

anz-gordonj7:
reopen if you want to.

PR #242, opened 2019-07-09
Anzdaddy/golangci lint

anzdaddy:
- Apply all golangci-lint linters
- Integrate golangci-lint into travis

To make it easier to review the impact per linter, I've created a commit for each linter that required code changes. Below, I've captured my thoughts on some of the linters. I'm inclined to keep them all in play for now, and keep an eye on whether they help or hinder in future.

### Several of the linters made me question their value:

1. **dupl:** required me to refactor a couple of listener_impl event handlers, which wasn't trivial and while I'm happy that the end result is a win, it's only just that, and it's not clear to me that the effort was worth it. I was nonetheless mildly impressed that it picked this up at all, given that the two blocks not only weren't identical but even used different types.
1. **gocritic:** I'm actually pretty sold on this one, however, it did cause a bit of a head-scratch with one if_else_else-to-switch scenario that had `break` statements in the original. In the end, however, I refactor the logic to avoid the chained if_else_else.
1. **gofmt:** The only caveat is that I can't get VSCode to apply the `-s` flag, since it's configured to use the experimental LSP engine, [gopls](https://github.com/golang/go/wiki/gopls). If I turn the experimental engine off, I could use `gofmt` with the `-s` parameter, but I would prefer `goreturns`, which doesn't have this feature. In any event, none of this has anything to do with the linter itself.
1. **scopelint:** It complains about all escaped loop variables, which could make it quite noisy and require a lot of unnecessary code to localise loop variables. However, on reflection, I've decided this could avoid some very sneaky bugs and is worth the extra code, even in cases where it's strictly not required.
1. **gochecknoinits:** Is avoiding `init()` functions recommended practice? It seems that people will simply use `init()` whenever they deem necessary and suppress the linter each time. I guess there's value in at least making people think about it before deciding this.
1. **gocyclo:** While I'm not a fan of cyclomatic complexity in general, I have to admit that it did yield more readable code in this instance.

### Some of them really didn't click for me:

1. **interfacer:** It feels strange to make some of the parameters more generic than the intended usage. I can't say that I have any objective objections. It just feels odd.
    - One objection that comes to mind is that future updates to the function might need to assume the more specific type. The YAGNI principle says to cross that bridge when you come to it, but by then the caller dependency on the more generic type might be much harder to peel back.
1. **unparam:** I'm unconvinced of the value of this one.
    - It complains that some functions are too generic because some parameters are always passed with the same value.
    - It also complains about unused variable names. This required removal of the `ctx` parameter name from most of the visitor methods, but I feel that this actually reduces readability.
1. **gochecknoglobals:** This felt like a waste of time. Most globals are there for good reason, and I just ended up suppressing this linter for every global and changing nothing.

@anz-bank/sysl-developers


sysl2/sysl/listener_impl.go
```diff
@@ -290,76 +289,58 @@ func (s *TreeShapeListener) EnterTypes(ctx *parser.TypesContext) {
...
 
-	if ctx.Size_spec() != nil {
+	if sizeSpec != nil {
 		if type1.GetPrimitive() != sysl.Type_NO_Primitive {
-			spec := ctx.Size_spec().(*parser.Size_specContext)
+			spec := sizeSpec.(*parser.Size_specContext)
 			type1.Constraint = makeTypeConstraint(type1.GetPrimitive(), spec)
 		}
 	}
+	return
```
sahejsingh: hmm learned a new thing today.

sysl2/sysl/parse.go
```diff
@@ -339,12 +339,12 @@ func infer_types(mod *sysl.Module, appName string) {
...
 			continue
 		}
-		infer_expr_type(mod, appName, view.Expr, true, 0)
+		inferExprType(mod, appName, view.Expr, true, 0)
 	}
 }
 
 func postProcess(mod *sysl.Module) {
-	var appNames []string
+	appNames := make([]string, 0, len(mod.Apps))
```
sahejsingh: hmm back to make() now? why?

sysl2/sysl/seqs.go
```diff
@@ -130,38 +132,61 @@ func DoConstructSequenceDiagrams(
 }
 
 // DoGenerateSequenceDiagrams generate sequence diagrams for the given model
-func DoGenerateSequenceDiagrams(stdout, stderr io.Writer, flags *flag.FlagSet, args []string) {
+func DoGenerateSequenceDiagrams(args []string) error {
```
sahejsingh: Keep stdout, stderr io.Writer for now
there are pull requests in pipeline that will use these variables.

sysl2/naive/grammarParser.go
```diff
@@ -163,13 +162,12 @@ func (p *parser) parse(g *sysl.Grammar, input int, val interface{}) (bool, int,
 				} else {
 					if matchCount < minCount {
 						return false, EOF, nil
-					} else if minCount == 0 && matchCount == 0 {
+					}
+					if matchCount == 0 {
```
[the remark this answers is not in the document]
anzdaddy: If minCount is 1 and matchCount is 0, then matchCount < minCount and it ends up in the first `if`.

sysl2/sysl/listener_impl.go
anzdaddy: Yeah, empty returns at the end of a function are not my cup of tea.

sysl2/sysl/parse.go
anzdaddy: It's recommended when the final size is known in advance. I was quite impressed that one of the linters picks up on these opportunities.

sysl2/sysl/seqs.go
anzdaddy: I'll hold off merging this PR until these other PRs land, and I'll incorporate their changes. We do have to bite the bullet at some point, though.

anzdaddy: Per out of band discussion, I've temporarily reinstated the parameters.

sysl2/naive/grammarParser.go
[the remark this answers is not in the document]
sahejsingh: yea. I had written if-else with the number line model, where minCount is at the origin/center and checking where all matchCount can be.
If you stare long enough you can find that minCount can only be 0 at this point. And that is why you can remove the minCount == 0 condition. Its provable but not super readable.
For clarity I like to keep this more explicit. Its easier to read that way.

PR #280, opened 2019-08-10
Remove temp test files

anzdaddy:
Change some tests to remove the temporary files they create.


anzdaddy:
> arguably it might be better to write files into a temp dir and then ensure the temp dir is cleaned.
> 
> but this is an incremental improvement over the status quo.

Agreed. Long term fix is to stub out the filesystem with an in-memory filesystem and have nothing written to disk.

PR #290, opened 2019-08-19
Bump version to 0.3.0

springwiz:


springwiz:
Last release with version 0.2.9 failed due to an issue in Pypi. Fix applied and bumped up the release version to 0.3.0.

"Update README.rst (#287)
  Pypi upload fails with rst rendering error. README is updated accordingly to fix it."

PR #292, opened 2019-08-23
Add version info into gosysl binary

### LINKED ISSUE #286: sysl -version should print the version of the binary
sysl -version should print the version of the binary

Use go compilers build meta data to support -version

kasunfdo:
Fixes #286 .

- Add version info into sysl binary so that `sysl --version` will show the binary version

@anz-bank/sysl-developers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,9 +57,9 @@ script:
 - if [[ ( "$TRAVIS_OS_NAME" == "linux" ) ]]; then gradle test -b test/java/build.gradle; fi
 - go test -coverprofile=coverage.txt -covermode=atomic github.com/anz-bank/sysl/...
 - npm test --prefix sysl2/sysl/sysl_js
-- GOOS=darwin GOARCH=amd64 go build -o gosysl/gosysl-darwin github.com/anz-bank/sysl/sysl2/sysl
-- GOOS=linux GOARCH=amd64 go build -o gosysl/gosysl-linux github.com/anz-bank/sysl/sysl2/sysl
-- go install github.com/anz-bank/sysl/sysl2/sysl
+- GOOS=darwin GOARCH=amd64 go build -o gosysl/gosysl-darwin -ldflags "-X main.Version=${TRAVIS_TAG:1}" github.com/anz-bank/sysl/sysl2/sysl
+- GOOS=linux GOARCH=amd64 go build -o gosysl/gosysl-linux -ldflags "-X main.Version=${TRAVIS_TAG:1}" github.com/anz-bank/sysl/sysl2/sysl
+- go install -ldflags "-X main.Version=${TRAVIS_TAG:1}" github.com/anz-bank/sysl/sysl2/sysl
 - bash test-gosysl.sh
 
 addons:
--- a/sysl2/sysl/sysl.go
+++ b/sysl2/sysl/sysl.go
@@ -15,12 +15,17 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
+// Version contains the sysl binary version
+//nolint:gochecknoglobals
+var Version = "unspecified"
+
 const debug string = "debug"
 
 // main3 is the real main function. It takes its output streams and command-line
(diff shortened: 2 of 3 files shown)


sysl2/sysl/sysl.go
```diff
@@ -13,12 +13,17 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
+// Version contains the sysl binary version
+// nolint:gochecknoglobals
```
anzdaddy: No whitespace before `nolint`. This is explicitly recommend by golangci-lint. https://github.com/golangci/golangci-lint#nolint
```go
//nolint:gochecknoglobals
```

.travis.yml
```diff
@@ -58,8 +58,8 @@ script:
 - gradle test -b test/java/build.gradle
 - go test -coverprofile=coverage.txt -covermode=atomic github.com/anz-bank/sysl/...
 - npm test --prefix sysl2/sysl/sysl_js
-- GOOS=darwin GOARCH=amd64 go build -o gosysl/gosysl-darwin github.com/anz-bank/sysl/sysl2/sysl
-- GOOS=linux GOARCH=amd64 go build -o gosysl/gosysl-linux github.com/anz-bank/sysl/sysl2/sysl
+- GOOS=darwin GOARCH=amd64 go build -o gosysl/gosysl-darwin -ldflags "-X main.Version=${TRAVIS_TAG:1}" github.com/anz-bank/sysl/sysl2/sysl
+- GOOS=linux GOARCH=amd64 go build -o gosysl/gosysl-linux -ldflags "-X main.Version=${TRAVIS_TAG:1}" github.com/anz-bank/sysl/sysl2/sysl
```
anzdaddy: How is `TRAVIS_TAG` computed?

anzdaddy: Defaulting to 1 doesn't look right. Better to use something like "unspecified".

anzdaddy: Needs a test, maybe add one to `test-gosysl.sh`. Since you can't predict the version number, just check that `--version` works.

anzdaddy: Would be good track additional info. This can be done later. #299

sysl2/sysl/sysl.go
```diff
@@ -13,12 +13,17 @@ import (
...
+var Version = "undefined"
```
anzdaddy: Prefer `"unspecified"`.

.travis.yml
kasunfdo: It is the same as git tag

kasunfdo: It is not defaulting to 1. As git tag it defined like `v0.3.0`, just removing `v` from tag.

kasunfdo: Agree. Although not included here, I have already added build date and commit id. Will address through #299.

sysl2/sysl/sysl.go
kasunfdo: IDE formats automatically by adding a space. 😀
Should turn that off

.travis.yml
kasunfdo: Done

sysl2/sysl/sysl.go
kasunfdo: Changed.

kasunfdo: Changed

sysl2/sysl/sysl.go
```diff
@@ -15,12 +15,17 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
+// Version contains the sysl binary version
+//nolint:gochecknoglobals
+var Version = "unspecified"
```
springwiz: Version should be a const.

kasunfdo: Although it sounds like a constant, cannot use constants as value should be reassigned during compile time

.travis.yml
anzdaddy: Ah, silly me. I mistook `:` for `:-`.

PR #304, opened 2019-08-28
Parallelise tests

anzdaddy:
Also introduces a pluggable filesystem, mainly for test isolation.

PR #306, opened 2019-09-02
Add coverage > 80% check to travis

anzdaddy:
Codecov has been un-required as a check in GitHub because its forced nonnegative delta was slowing us down unduly. This provides an alternative measure to ensure coverage doesn't regress significantly.

This is mostly a delta on #304. Only the final commit is significant.


.travis.yml
```diff
@@ -55,7 +55,7 @@ script:
 - pytest
 - pytest test/e2e --syslexe=${SYSL_PYTHON_DIR}sysl --reljamexe=${SYSL_PYTHON_DIR}reljam
 - if [[ ( "$TRAVIS_OS_NAME" == "linux" ) ]]; then gradle test -b test/java/build.gradle; fi
-- go test -coverprofile=coverage.txt -covermode=atomic github.com/anz-bank/sysl/...
+- ./check-coverage.sh
```
kasunfdo: Missing arg

anzdaddy: You beat me to it!

PR #307, opened 2019-09-04
GitHub Actions Go build

anzdaddy:
Keeping the other CI process in place for now.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,101 @@
+name: Go
+on: [push]
+jobs:
+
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -0,0 +1,101 @@
+name: Go
+on: [push]
+jobs:
+
+  build-and-test-linux:
```
springwiz: Don't see a Windows build?

anzdaddy: Not yet. We can do that in a later PR.

PR #308, opened 2019-09-05
Inject rich context info into go-sysl binary

### LINKED ISSUE #299: Inject rich context info into binaries
#292 adds version info to the sysl binary. Would be good to also track:
- git commit id
- git branch.
- build timestamp (`-ldflags "-X main.BuildTimestamp=$(date -u +'%Y-%m-%dT%H:%M:%SZ')"`).
- $GOOS, $GOARCH and Go version

Not sure if this is easy, or even possible, but the following output would be ideal:

```bash
$ sysl --version
1.2
$ sysl --version -v
sysl version 1.2
git master 7537f593757c20372ab9144cf7ac8fcc23528f17
built 2019-08-26T21:04:34Z
go version go1.12.9 darwin/amd64
```

Other info that caught my attention at https://docs.travis-ci.com/user/environment-variables/:

- TRAVIS_BUILD_NUMBER
- TRAVIS_DIST
- TRAVIS_OS_NAME
- TRAVIS_OSX_IMAGE

kasunfdo:
Fixes #299 .

Following info is injected into the binary

- Version  
- GitCommit
- BuildDate
- GoOS     
- GoArch   
- GoVersion
- BuildOS  

Use `sysl --info` to view above info


.travis.yml
```diff
@@ -57,9 +62,9 @@ script:
...
-- GOOS=linux GOARCH=amd64 go build -o gosysl/gosysl-linux -ldflags "-X main.Version=${TRAVIS_TAG:1}" github.com/anz-bank/sysl/sysl2/sysl
-- go install -ldflags "-X main.Version=${TRAVIS_TAG:1}" github.com/anz-bank/sysl/sysl2/sysl
+- GOOS=${GOOS_DARWIN} GOARCH=${GOARCH} go build -o gosysl/gosysl-darwin -ldflags "-X \"main.Version=${TRAVIS_TAG:1}\" -X \"main.GitCommit=${TRAVIS_COMMIT}\" -X \"main.BuildDate=${BUILD_DATE}\" -X \"main.GoArch=${GOARCH}\" -X \"main.GoOS=${GOOS_DARWIN}\" -X \"main.GoVersion=${GO_VERSION}\" -X \"main.TravisBuildNum=${TRAVIS_BUILD_NUMBER}\" -X \"main.TravisDist=${TRAVIS_DIST}\" -X \"main.TravisOS=${TRAVIS_OS_NAME}\"" -v github.com/anz-bank/sysl/sysl2/sysl
+- GOOS=${GOOS_LINUX}  GOARCH=${GOARCH} go build -o gosysl/gosysl-linux  -ldflags "-X \"main.Version=${TRAVIS_TAG:1}\" -X \"main.GitCommit=${TRAVIS_COMMIT}\" -X \"main.BuildDate=${BUILD_DATE}\" -X \"main.GoArch=${GOARCH}\" -X \"main.GoOS=${GOOS_LINUX}\"  -X \"main.GoVersion=${GO_VERSION}\" -X \"main.TravisBuildNum=${TRAVIS_BUILD_NUMBER}\" -X \"main.TravisDist=${TRAVIS_DIST}\" -X \"main.TravisOS=${TRAVIS_OS_NAME}\"" -v github.com/anz-bank/sysl/sysl2/sysl
```
anzdaddy: This should be a script, and the envvars should just be parameters to the script.

.travis.yml
```diff
@@ -57,9 +62,9 @@ script:
...
+- go install -ldflags "-X \"main.Version=${TRAVIS_TAG:1}\" -X \"main.GitCommit=${TRAVIS_COMMIT}\" -X \"main.BuildDate=${BUILD_DATE}\" -X \"main.GoArch=`uname -m`\" -X \"main.GoOS=`uname -s`\" -X \"main.GoVersion=${GO_VERSION}\" -X \"main.TravisBuildNum=${TRAVIS_BUILD_NUMBER}\" -X \"main.TravisDist=${TRAVIS_DIST}\" -X \"main.TravisOS=${TRAVIS_OS_NAME}\"" github.com/anz-bank/sysl/sysl2/sysl
```
anzdaddy: Does `uname` output match the expected GOOS and GOARCH values?

kasunfdo: Nope. I used it in that way as there's no arg is used for GOOS and GOARCH

.travis.yml
anzdaddy: That is, there should be no env vars, just parameters.

.travis.yml
anzdaddy: It should match what the values would have been, if specified. You can extract this information from `go version` instead.

kasunfdo: Thank you! 💡💡

scripts/build.sh
```diff
@@ -0,0 +1,70 @@
...
+    usage
+    exit 1
+fi
+
+FLAGS="\
+    -X \"main.Version=${VERSION}\" \
+    -X \"main.GitCommit=${COMMIT}\" \
+    -X \"main.BuildDate=${BUILD_DATE}\" \
+    -X \"main.GoOS=${GOOS}\" \
+    -X \"main.GoArch=${GOARCH}\" \
```
anzdaddy: Sorry, I should have realised before, but GOOS and GOARCH are already baked into the binary. https://golang.org/pkg/runtime/#pkg-constants

scripts/build.sh
```diff
@@ -0,0 +1,73 @@
...
+fi
+if [[ ${CMD} == "build" && (-z ${GOOS} || -z ${GOARCH}) ]]; then
+  echo "GO OS and/or GO Arch is not specified. Default go build command will be used"
+fi
+
+FLAGS="\
+    -X \"main.Version=${VERSION}\" \
+    -X \"main.GitCommit=${COMMIT}\" \
+    -X \"main.BuildDate=${BUILD_DATE}\" \
+    -X \"main.GoVersion=${GO_VERSION}\" \
```
anzdaddy: `GO_VERSION` is also not required. https://golang.org/pkg/runtime/#Version

sysl2/sysl/sysl.go
```diff
@@ -106,3 +114,16 @@ func main() {
...
 }
+
+//nolint:unparam
+func showBinaryInfo(_ *kingpin.ParseContext) error {
+	fmt.Printf("Version    : %s\n", Version)
+	fmt.Printf("Commit ID  : %s\n", GitCommit)
+	fmt.Printf("Build date : %s\n", BuildDate)
+	fmt.Printf("GoOS       : %s\n", runtime.GOOS)
+	fmt.Printf("GoArch     : %s\n", runtime.GOARCH)
+	fmt.Printf("Golang     : %s\n", GoVersion)
```
anzdaddy: `s/Golang/Go version/`

Btw, the language is called Go, not Golang.

sysl2/sysl/sysl.go
```diff
@@ -106,3 +114,16 @@ func main() {
 		os.Exit(rc)
 	}
 }
+
+//nolint:unparam
+func showBinaryInfo(_ *kingpin.ParseContext) error {
+	fmt.Printf("Version    : %s\n", Version)
+	fmt.Printf("Commit ID  : %s\n", GitCommit)
+	fmt.Printf("Build date : %s\n", BuildDate)
+	fmt.Printf("GoOS       : %s\n", runtime.GOOS)
```
anzdaddy: `s/GoOS/GOOS/`

sysl2/sysl/sysl.go
```diff
@@ -106,3 +114,16 @@ func main() {
...
+	fmt.Printf("GoArch     : %s\n", runtime.GOARCH)
```
anzdaddy: `s/GoArch/GOARCH/`

.github/workflows/go-darwin.yml
```diff
@@ -47,12 +47,12 @@ jobs:
         export GOBIN=$PWD/bin
         mkdir $GOBIN
         GIT_TAG="$(echo $GITHUB_REF | cut -d'/' -f3)"
-        go install -ldflags "-X main.Version=${GIT_TAG:1}" ./sysl2/sysl
+        bash scripts/build.sh install ${GIT_TAG:1} ${GITHUB_SHA} ${OSTYPE}
```
anzdaddy: `bash` prefix is redundant. `scripts/build.sh …` should suffice. Is the script `chmod +x`-ed?

Ditto everywhere else `bash` is invoked.

sysl2/sysl/sysl.go
```diff
@@ -106,3 +114,16 @@ func main() {
...
+	fmt.Printf("Golang     : %s\n", GoVersion)
+	fmt.Printf("BuildOS    : %s\n", BuildOS)
```
anzdaddy: `s/BuildOS/Build OS/`

sysl2/sysl/sysl.go
```diff
@@ -106,3 +114,16 @@ func main() {
 		os.Exit(rc)
 	}
 }
+
+//nolint:unparam
+func showBinaryInfo(_ *kingpin.ParseContext) error {
+	fmt.Printf("Version    : %s\n", Version)
+	fmt.Printf("Commit ID  : %s\n", GitCommit)
```
anzdaddy: OK for now, but #338

sysl2/sysl/sysl.go
```diff
@@ -106,3 +114,16 @@ func main() {
...
+	fmt.Printf("Build date : %s\n", BuildDate)
+	fmt.Printf("GoOS       : %s\n", runtime.GOOS)
+	fmt.Printf("GoArch     : %s\n", runtime.GOARCH)
+	fmt.Printf("Golang     : %s\n", GoVersion)
+	fmt.Printf("BuildOS    : %s\n", BuildOS)
+	os.Exit(0)
```
anzdaddy: This shouldn't be needed. I can see that it's being used to hack the command-line parser. If you really want to do that, supply a lambda that calls showBinaryInfo and then exits. But this really should be implemented as a subcommand, not an option hack.

.github/workflows/release.yml
```diff
@@ -44,13 +44,14 @@ jobs:
     - name: Test and Build
       run: |
         GIT_TAG="$(echo $GITHUB_REF | cut -d'/' -f3)"
-        ./check-coverage.sh 80
-        bash test-pysysl.sh
+        bash scripts/check-coverage.sh 80
+        bash scripts/test-pysysl.sh
+        gradle test -b test/java/build.gradle
```
anzdaddy: Could we merely depend on the other workflows rather than rerunning test?

scripts/build.sh
```diff
@@ -0,0 +1,73 @@
...
+  usage
+  exit 1
+fi
+if [[ ${CMD} == "build" && -z ${OUT} ]]; then
+  echo "Output is not specified"
+  usage
+  exit 1
+fi
+if [[ ${CMD} == "build" && (-z ${GOOS} || -z ${GOARCH}) ]]; then
+  echo "GO OS and/or GO Arch is not specified. Default go build command will be used"
```
anzdaddy: `s/GO OS/GOOS/` `s/GO Arch/GOARCH/`

scripts/build.sh
```diff
@@ -0,0 +1,73 @@
...
+fi
+
+FLAGS="\
+    -X \"main.Version=${VERSION}\" \
```
anzdaddy: Use `'` instead of `\"`.

sysl2/sysl/sysl_test.go
```diff
@@ -599,3 +601,19 @@ func TestMain2WithDataMultipleRelationships(t *testing.T) {
...
+		panic("os.Exit called")
+	}
+	patch := monkey.Patch(os.Exit, fakeExit)
+	defer patch.Unpatch()
+	assert.PanicsWithValue(
+		t,
+		"os.Exit called",
+		func() {
+			err := showBinaryInfo(nil)
+			assert.NoError(t, err)
```
anzdaddy: If `showBinaryInfo` panics, you should test that the flow never arrives here:

```go
t.Error("Shouldn't get here.")
```

But this is moot if you reimplement this feature as a subcommand, as recommended elsewhere.

sysl2/sysl/sysl_test.go
```diff
@@ -599,3 +601,19 @@ func TestMain2WithDataMultipleRelationships(t *testing.T) {
 		fs, logger, main3)
 	testutil.AssertFsHasExactly(t, memFs, "/Relational-Model.png", "/Object-Model.png")
 }
+
+func TestShowBinaryInfo(t *testing.T) {
+	fakeExit := func(int) {
+		panic("os.Exit called")
+	}
+	patch := monkey.Patch(os.Exit, fakeExit)
```
anzdaddy: This is why hacks like calling `os.Exit` from an option handler are not a good idea. :-)

.github/workflows/release.yml
kasunfdo: The reason I did this is GHAs run in parallel and I couldn't find a way to wait for other runs are completed.
[jobs.<job_id>.needs](https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idneeds) works only if all the jobs are in same action script.

I did not want to publish release artifacts while other runs are failing. That's why I included tests in release script. During a release, only this action script will be executed.

sysl2/sysl/sysl.go
kasunfdo: Agree. Changed implementation to use a command instead of an option.
`> sysl info`

.github/workflows/go-darwin.yml
kasunfdo: Yes. All scripts are executable.

scripts/build.sh
```diff
@@ -0,0 +1,70 @@
...
+    echo "Invalid command"
+    usage
+    exit 1
+elif [[ -z ${VERSION} ]]; then
+    echo "Version is not specified"
+    usage
+    exit 1
+elif ! [[ ${VERSION} =~ ^[[:digit:]+\\.[:digit:]+\\.[:digit:]]+$ ]]; then
+    echo "Version is invalid. Binary version will be empty"
+    VERSION=""
```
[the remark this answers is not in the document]
kasunfdo: IMO current implementation is more readable as it is. I you really insist, I will make the change.

scripts/build.sh
```diff
@@ -0,0 +1,70 @@
...
+}
+
+if [[ -z ${CMD} ]]; then
+    echo "Command is not specified"
+    usage
+    exit 1
+elif [[ ${CMD} != "build" && ${CMD} != "install"  ]]; then
+    echo "Invalid command"
+    usage
+    exit 1
```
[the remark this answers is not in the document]
kasunfdo: IMO current implementation is more readable as it is. I you really insist, I will make the change.

sysl2/sysl/cmd_binaryinfo.go
```diff
@@ -0,0 +1,29 @@
+package main
```
golangcibot: /go/pkg/mod/github.com/stretchr/testify@v1.4.0/assert/assertions.go:174:18: DecodeRuneInString not declared by package utf8 (from `typecheck`)

PR #335, opened 2019-09-17
Feature/move more commands

anz-gordonj7:
More code reorg to reduce duplication.

Moves the sd and gen commands into the new style. Minor breaking change on the gen command:

* Command name is now 'codegen' with the 'gen' alias
* --root-model is removed, use the top-level --root flag
* --model is removed, MODULE is the argument after the end of thee command flags

@anz-bank/sysl-developers


sysl2/sysl/cmd_runner.go
```diff
@@ -63,8 +63,35 @@ func (r *Runner) Configure(app *kingpin.Application) error {
...
+		}
+		if errorMsg.Len() > 0 {
+			return errors.New(errorMsg.String())
+		}
+		return nil
+
+	}
+
+	cmd.PreAction(fn)
+}
```
springwiz: Kingpin provides OOB validation hooks at the application and the comand level. These hooks are automaticalIy invoked when the command line is parsed. I suggest using them instead.  
type CmdClauseValidator func(*CmdClause) error

anz-gordonj7: I tried that, cant remember why it didnt work though, will try it again.

anz-gordonj7: @springwiz yeah, just checked again, validate appaears to happen at the wrong point in the parsing, PreAction works better.

PR #342, opened 2019-09-20
Implement count on map

### LINKED ISSUE #341: Implement count on map
Currently, count functions accepts only sets and lists. Need to use count function to get number of keys in a map

### Expected behaviour
Sysl statement: `keyCount = map count`
Value of keyCount should be the number of keys in map

kasunfdo:
Fixes #341

### DIFF
--- a/sysl2/sysl/eval/exprEval.go
+++ b/sysl2/sysl/eval/exprEval.go
@@ -176,12 +176,16 @@ func evalCall(txApp *sysl.Application, assign Scope, x *sysl.Expr_Call_) *sysl.V
 		switch x.Call.Func[1:] {
 		case "count":
 			argExpr := Eval(txApp, assign, x.Call.Arg[0])
-			if argExpr.GetList() != nil {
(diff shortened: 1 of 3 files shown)


sysl2/sysl/eval/exprEval.go
```diff
@@ -176,12 +176,16 @@ func evalCall(txApp *sysl.Application, assign Scope, x *sysl.Expr_Call_) *sysl.V
 		switch x.Call.Func[1:] {
 		case "count":
 			argExpr := Eval(txApp, assign, x.Call.Arg[0])
-			if argExpr.GetList() != nil {
-				return MakeValueI64(int64(len(argExpr.GetList().Value)))
-			} else if argExpr.GetSet() != nil {
-				return MakeValueI64(int64(len(argExpr.GetSet().Value)))
+			switch t := argExpr.Value.(type) {
```
springwiz: How do we plan to test this change?

PR #344, opened 2019-09-23
Add command line argument to read app name

### LINKED ISSUE #343: Add command line argument to read app name
In code generation, name of app is chosen by iterating over the map of apps and selecting the first app name.
As the result, when there are multiple apps defined or imported in a sysl model, generated output becomes unpredictable.

### Expected behaviour
When code is generated for foo.sysl, there exists code only for foo.sysl

### Actual behaviour
When code is generated for foo.sysl, sometimes code is being generated for bar.sysl as well

Generate code for a sysl file with multiple apps

eg:

foo.sysl
```python
import Bar

Foo:
  !type Todo:
    id <: int

  /foo:
    /{id<:int}:
      GET:
        Bar <- GET /bar/{id}
        return Todo
```

bar.sysl
```
Bar:
  !type Bar:
    result <: string

  /bar/{id<:int}:
    GET:
      return Bar
```

kasunfdo:
Fixes #343 .

- Use `--app` flag to provide name of the model app


sysl2/sysl/types.go
```diff
@@ -47,6 +47,7 @@ type CmdContextParamCodegen struct {
 	transform     string
 	grammar       string
 	start         string
+	appName       string
```
golangcibot: `appName` is unused (from `structcheck`)

sysl2/sysl/cmd_codegen.go
```diff
@@ -29,7 +30,7 @@ func (p *codegenCmd) Configure(app *kingpin.Application) *kingpin.CmdClause {
 
 func (p *codegenCmd) Execute(args ExecuteArgs) error {
 
-	output, err := GenerateCode(&p.CmdContextParamCodegen, args.Module, args.ModuleAppName, args.Filesystem, args.Logger)
+	output, err := GenerateCode(&p.CmdContextParamCodegen, args.Module, p.appName, args.Filesystem, args.Logger)
```
anzdaddy: What happens to the `ModuleAppName` field? Can it be removed?

sysl2/sysl/cmd_codegen.go
```diff
@@ -21,6 +21,7 @@ func (p *codegenCmd) Configure(app *kingpin.Application) *kingpin.CmdClause {
 		Default(".").StringVar(&p.rootTransform)
 	cmd.Flag("transform", "grammar.g").Required().StringVar(&p.transform)
 	cmd.Flag("grammar", "grammar.g").Required().StringVar(&p.grammar)
+	cmd.Flag("app", "name of the app").Required().StringVar(&p.appName)
```
anz-gordonj7: please use a more descriptive flag name. Also does this flag makes sense on any other commands?

springwiz: I think the help string could be a little more descriptive.

sysl2/sysl/cmd_codegen.go
kasunfdo: Opened PR by a mistake.... 😞
Yes ModuleAppName is not required. Will remove it

sysl2/sysl/cmd_codegen.go
kasunfdo: Nope. It's only required for codegen.

sysl2/sysl/cmd_codegen.go
```diff
@@ -19,8 +20,9 @@ func (p *codegenCmd) Configure(app *kingpin.Application) *kingpin.CmdClause {
 	cmd.Flag("root-transform",
 		"sysl root directory for input transform file (default: .)").
 		Default(".").StringVar(&p.rootTransform)
-	cmd.Flag("transform", "grammar.g").Required().StringVar(&p.transform)
-	cmd.Flag("grammar", "grammar.g").Required().StringVar(&p.grammar)
+	cmd.Flag("transform", "path to transform file from the root transform directory").Required().StringVar(&p.transform)
+	cmd.Flag("grammar", "path to grammar file").Required().StringVar(&p.grammar)
+	cmd.Flag("app", "name of the sysl app defined in sysl model. if there are multiple apps defined in sysl model, code will be generated only for the given app").Required().StringVar(&p.appName)
```
golangcibot: line is 192 characters (from `lll`)

sysl2/sysl/cmd_codegen.go
kasunfdo: Updated description

kasunfdo:
> Why is this required? It could be made that if the flag is missing every app in the module can get generated with repeated calls to GenerateCode()? Also, really would like a better flag name

Currently, transform is not written to handle code generation for multiple apps. If it is done, there will be naming conflicts in generated files. As generated wiles will be written in output dir, it might contain files with different package names.

I understand your concern. Until that feature is available, it would be better to add the flag so that there will not be surprises in generated code.

PR #345, opened 2019-09-24
Initial work to import swagger -> sysl via go

anz-gordonj7:
Testcases came from the python output, more tests needed for full swagger spec coverage though`

All the test sysl files are unmodified except SWAGGER_OBJECT_WITH_NO_PROPERTIES.sysl which has a empty newline removed to match the go output.

There is one test which fails infrequently due to map key ordering which I'm going to fix this morning but thought it was worth getting this out to start working through review issues.

@anz-bank/sysl-developers


sysl2/sysl/cmd_importswagger.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return err
+	}
+
+	output, err := swagger.LoadSwaggerText(p.OutputData, string(data))
+	if err != nil {
+		return err
+	}
+
+	ioutil.WriteFile(p.outfile, []byte(output), 0644)
```
golangcibot: Error return value of `ioutil.WriteFile` is not checked (from `errcheck`)

sysl2/sysl/swagger/import.go
```diff
@@ -0,0 +1,319 @@
...
+	si.writeLines(fmt.Sprintf("%s %s [package=%s]:", si.args.AppName, quote(title), quote(si.args.Package)))
+	si.ind.Push()
+
+	return nil
+}
+
+func buildQueryString(params []Param) string {
+	query := ""
+	if len(params) > 0 {
+		parts := []string{}
```
kasunfdo: nit: nil slice is preferred.
`var parts []string`

sysl2/sysl/swagger/import.go
```diff
@@ -0,0 +1,319 @@
...
+	return query
+}
+
+func buildRequestBodyString(params []Param) string {
+	body := ""
+	if len(params) > 0 {
+		sort.SliceStable(params, func(i, j int) bool {
+			return strings.Compare(params[i].Name, params[j].Name) < 0
+		})
+		parts := []string{}
```
kasunfdo: nit: nil slice is preferred.
`var parts []string`

sysl2/sysl/swagger/import.go
```diff
@@ -0,0 +1,319 @@
...
+	return body
+}
+
+func buildRequestHeadersString(params []Param) string {
+	headers := ""
+	if len(params) > 0 {
+		sort.SliceStable(params, func(i, j int) bool {
+			return strings.Compare(params[i].Name, params[j].Name) < 0
+		})
+		parts := []string{}
```
kasunfdo: nit: nil slice is preferred.
`var parts []string`

anz-gordonj7: Fixed in follow up branch.

sysl2/sysl/swagger/import.go
anz-gordonj7: Fixed in follow up branch.

sysl2/sysl/swagger/types.go
```diff
@@ -0,0 +1,207 @@
...
+			"float":  "float",
+		},
+	}
+
+	if formatMap, ok := conversions[typeName]; ok {
+		if result, ok := formatMap[format]; ok {
+			return result
+		}
+	}
+
```
kasunfdo: It would be great if a descriptive log is shown for failure scenarios. It would help to find spec issues and potential scenarios that can be handled by importer

sysl2/sysl/swagger/import.go
```diff
@@ -0,0 +1,319 @@
...
+		for statusCode, response := range endpoint.Responses.StatusCodeResponses {
+			if schema := response.Schema; schema != nil {
+				outs = append(outs, findReferencedType(*schema))
+			} else {
+				outs = append(outs, fmt.Sprintf("%d", statusCode))
+			}
+
+		}
+		if endpoint.Responses.Extensions != nil {
+			println("x-* responses not implemented")
```
kasunfdo: It would be great if a descriptive log is shown instead of println

sysl2/sysl/swagger/import.go
```diff
@@ -0,0 +1,319 @@
...
+			for key := range endpoint.Responses.Extensions {
+				outs = append(outs, key)
+			}
+		}
+		if endpoint.Responses.Default != nil {
+			println("default responses not implemented")
```
kasunfdo: It would be great if a descriptive log is shown instead of println

sysl2/sysl/swagger/parameters.go
```diff
@@ -0,0 +1,106 @@
...
+				return p
+			}
+			log.Panicf("referenced parameter %s unknown", refParamName)
+		}
+	} else if paramTypeName != "string" {
+		paramTypeName = mapSwaggerTypeAndFormatToType(param.Type, param.Format)
+	}
+	ptype, found := types.Find(paramTypeName)
+	if !found {
+		fmt.Printf("WARNING: Type %s unknown in param %s\n", paramTypeName, param.Name)
```
kasunfdo: Please use logging instead of printf

sysl2/sysl/swagger/types.go
```diff
@@ -0,0 +1,207 @@
...
+		return strings.Compare(keys[i], keys[j]) < 0
+	})
+	for _, k := range keys {
+		t.Properties = append(t.Properties, fields[k])
+	}
+}
+
+func mapSwaggerTypeAndFormatToType(typeName, format string) string {
+	if format != "" && !contains(format, swaggerFormats) {
+		panic("unknown format")
```
kasunfdo: Log and terminate instead of panic. LogFatal?

sysl2/sysl/swagger/types.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	}
+
+	return typeName
+}
+
+func InitTypes(doc *spec.Swagger) TypeList {
+	defs := TypeList{}
+	for name, definition := range doc.Definitions {
+		if _, found := defs.Find(name); !found {
```
kasunfdo: This gave me a feeling that `Find(...)` will never return true until I check the implementation of `Find(...)`.
I would suggest to use `Find` method to find within the `TypeList` and implement `FindElseWhere` separately.

kasunfdo: Or make it a function instead of method.

sysl2/sysl/swagger/tests/EXAMPLE_SWAGGER_SPEC_WITH_ENDPOINT_PATH_WITH_DEFAULT_RESPONSE.sysl
```diff
@@ -0,0 +1,21 @@
...
+    @host = "goat.example.com"
+    @description =:
+        | No description.
+
+    /api/v1:
+
+        /goat/status:
+            GET:
+                | Check goat status
+                return default
```
kasunfdo: IMO this should be just `return`

sysl2/sysl/swagger/types.go
anz-gordonj7: Can you elaborate? I'm not sure I understand your reasoning. its analogous of doing a map lookup.

sysl2/sysl/swagger/tests/EXAMPLE_SWAGGER_SPEC_WITH_ENDPOINT_PATH_WITH_DEFAULT_RESPONSE.sysl
anz-gordonj7: This is exactly what the python generated... LEts turn this into a ticket.

sysl2/sysl/swagger/import.go
anz-gordonj7: fixed all 3

sysl2/sysl/swagger/types.go
anz-gordonj7: done

sysl2/sysl/swagger/import.go
anz-gordonj7: done

sysl2/sysl/swagger/import.go
anz-gordonj7: done

sysl2/sysl/swagger/parameters.go
anz-gordonj7: done

sysl2/sysl/swagger/types.go
anz-gordonj7: done

sysl2/sysl/swagger/import.go
```diff
@@ -0,0 +1,319 @@
...
+	args    OutputData
+
+	types        TypeList
+	endpoints    map[string][]Endpoint
+	globalParams Parameters
+}
+
+func (si *swaggerImporter) mustWrite(s string) {
+	if _, err := si.Writer.Write([]byte(s)); err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
anz-gordonj7: done

sysl2/sysl/swagger/types.go
anz-gordonj7: Simplified it a little but, but this Find() is used everywhere so I'd rather not have to make 2 func calls in each place it is used.

PR #347, opened 2019-09-25
Define rest api status code to return type mapping

springwiz:
Fixes # .

- Define rest api status code to return type mapping

@anz-bank/sysl-developers


sysl2/sysl/eval/ValueUtil.go
```diff
@@ -262,12 +266,23 @@ func endpointToValue(e *sysl.Endpoint) *sysl.Value {
 
 	stmtsList := MakeValueList()
 	for _, stmt := range e.Stmt {
-		if stmt.GetRet() != nil {
-			AddItemToValueMap(m, "ret", stmtToValue(stmt))
-		} else {
+		switch {
```
kasunfdo: Handle `else if` as well.

I prefer type switches as follows.

```go
		switch s := stmt.Stmt.(type) {
		case *sysl.Statement_Ret:
			retValues = strings.Split(s.Ret.GetPayload(), " <: ")
		case *sysl.Statement_Cond:
			retValues = strings.Split(s.Cond.GetStmt()[0].GetRet().GetPayload(), " <: ")
		case *sysl.Statement_Group:
			if s.Group.GetTitle() == "else" || strings.Contains(s.Group.GetTitle(), "else if") {
				retValues = strings.Split(s.Group.GetStmt()[0].GetRet().GetPayload(), " <: ")
			}else {
				// Error/Warn
			}
		default:
			AppendItemToValueList(stmtsList.GetList(), stmtToValue(stmt))
		}
```

sysl2/sysl/eval/ValueUtil.go
```diff
@@ -236,6 +239,8 @@ func endpointToValue(e *sysl.Endpoint) *sysl.Value {
 	AddItemToValueMap(m, "attrs", attrsToValueMap(e.Attrs))
 	AddItemToValueMap(m, "is_rest", MakeValueBool(e.RestParams != nil))
 	AddItemToValueMap(m, "is_pubsub", MakeValueBool(e.IsPubsub))
+	retTypes := MakeValueMap()
+	retValues := []string{}
```
kasunfdo: Sorry, I missed this earlier. Please use nil slice declaration instead of empty slice.
`var retValues []string`

springwiz: ok

PR #348, opened 2019-09-25
Various import fixes and added test cases:

anz-gordonj7:
array handling
alias handling
don't sort header fields

Fixes # .

-
-

@anz-bank/sysl-developers


sysl2/sysl/swagger/tests/SWAGGER_WITH_ENUMS.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+definitions:
+
+  Attachment:
+    additionalProperties: false
+    properties:
+      data:
+        maxLength: 10000
+        type: string
+      type:
```
kasunfdo: Nice to have alias for this as well.
```python
type <: AttachementType
```

```python
!alias AttachementType:
        string
```

anz-gordonj7: yeah, I'm having a quick look and I'm not sure how to add that in.

I think the way I'm handling alias/array types needs to be reworked to support this.
This ties into my thinking also that `findReferencedType()` is bad.

It really depends on how much we expect to want to extend this to decide if its worth fixing. I think it would be good to dump the enum values as a doc string for the type until sysl gets actual enum support. either way, defer this.

PR #363, opened 2019-10-11
Allow hypen in query string variable names if it is prefixed with @

### LINKED ISSUE #362: hyphen in query variable name is rejected by the parser
### Expected behaviour

URL line line `POST ?goat-id=string:` should be legal

### Actual behaviour

Parser complains with an unhelpful message, removing the hyphen fixes the issue (but wont generate correct output as far as the API is defined)

### Possible solution

With my minimal understanding of the lexer/parser we can solve this by requiring the '@' symbol to prefix the query var name if the name has any special characters (i.e the hypen).

This change will likely require some codegen changes to be supported, and I'm not sure if this will break any real endpoints (I see that @ is reserved in the URI so it should be safe)

@anz-bank/sysl-developers thoughts?

anz-gordonj7:
Fixes #362 

 - Change the grammar to allow URI query strings to contain hypens in vairable names as long as the variable is prefixed with a `@`

@anz-bank/sysl-developers


sysl2/sysl/parse/listener_impl.go
```diff
@@ -1100,7 +1100,13 @@ func (s *TreeShapeListener) ExitDocumentation_stmts(*parser.Documentation_stmtsC
 
 // EnterQuery_var is called when production query_var is entered.
 func (s *TreeShapeListener) EnterQuery_var(ctx *parser.Query_varContext) {
-	var_name := ctx.Name().GetText()
+
```
anzdaddy: Doesn't the new linter complain about these leading empty lines?

gkanz:
The query params come through in the SYSL as something like `"?cust-nums=string&summary-mode=bool?"`. How would the `@` be used?

sysl2/sysl/parse/listener_impl.go
anz-gordonj7: I think that file is excluded becuase it has a ton of bad lines already

anz-gordonj7:
@gkanz 

> The query params come through in the SYSL as something like `"?cust-nums=string&summary-mode=bool?"`. How would the `@` be used? `"?@cust-nums=string&@summary-mode=bool?"`?

Correct

> Will this impact downstream consumers like the VS Code plugin? 

IT shouldnt, the vscode plugin uses the sysl validate command, so it will need to be updated, and the way it is implemented, the @ symbol is not added to the sysl AST(?)

> Is this impacting _all_ query params? That is: `"?@custNums=string&@summaryMode=bool?"`?

The @ is only needed for params with - in the name, it isnt reuired otherwise.

To be honest, I'm not happy with this workaround, and want to investigate how the parser works more to fix this properly. (I'm not done antlr4 before). the problem is that ANTLR4 works in two passes, the first tokenises the input, the second walks the tokens, so the hyphen isnt being captured by the tokeniser in the right token, if I can get the token stream I can see which token is breaking it and fix it correctly. (Or maybe fix it in the parser step)... dunno

gkanz:
I've made a change in the python swagger importer that uses a similar approach to how header parameters with hyphens are handled. It is still ugly but as consistent as I could make it. In the above example the following attribute would be added to the end point: `[queryParamNames=[["custNums","cust-nums"],["summaryMode","summary-mode"]]]`. For header params the current solution is to have a `[name="My-Header-Id"]` attribute. The above approach also allows for other attributes to be attached to the query params (formatting directives might be an example).

PR #365, opened 2019-10-16
Start tightening up the grammar to not allow free-text everywhere

### LINKED ISSUE #362: hyphen in query variable name is rejected by the parser

anz-gordonj7:
Fixes #362. This grammar change means that any useage of "name_str" in the REST HTTP line will not allow free-text
	    which will then allow the usual identifiers (letters/numebrs and hypens) without allowing everything else

* I'd also like to look into removing the TEXT_LINE regex completly if that is possible which will simplify things slightly.

@anz-bank/sysl-developers


anz-gordonj7:
@anz-bank/sysl-developers the second commit adds a feature toggle flag to not break existing files, if we are OK with this feature I'll add documentation and warnings (and happy to change the keyworkd name)

PR #368, opened 2019-10-18
Change import statement to allow direct importing swagger files

anz-gordonj7:
Spec:
`import FILENAME [as [PACKAGE_NAME.]APP_NAME] [~MODE] [# comments...]`

where:
	FILENAME - path to the file to import (i.e backend.yaml)

The following is required if importing non-sysl files
	PACKAGE_NAME (optional) - Package name of the generated sysl
	APP_NAME - Name of the application of the generated sysl

	MODE - import mode, empty is guess based on the extension,
		currently supported:
			~swagger

Fixes # .

-
-

@anz-bank/sysl-developers


anz-gordonj7:
@anzdaddy If you're happy with this approve and merge please, no need to wait till wednesday for merge if green ticks.

sysl2/sysl/parse/tests/foreign_import_swagger.sysl
```diff
@@ -0,0 +1,20 @@
+import foreign_import_swagger.yaml as pck.app ~swagger
```
anzdaddy: What's the significance of `pck.app`? Should the ref below be `…depth=pck.app.SimpleObj&…` or similar? Referencing syntax and semantics is something we need to clean up sooner rather than later.

anz-gordonj7: the import code needs an application name (at least for swagger imports anyway), so in that line `app` would be the application name for use later. `pck` would be the full package name, maybe the example should be something better like `com.anz.dependancy.foo` where `foo` ends up being the app name as the rest is the package name.

anz-gordonj7:
@nofun97 fyi

PR #376, opened 2019-10-21
Make the export command generic

springwiz:
Fixes # .

- Make export command generic, swagger export one of the targets
- Allows to choose format(yaml/json), yaml default

@anz-bank/sysl-developers


springwiz:
> I don’t see any parallelisation of unit tests. Probably better to keep that split out from this PR anyway. Please comment and update the description if necessary.

sysl2/sysl/exporter/swagger_exporter_test.go -> TestExportAll()
Has multiple t.run() calls now. Earlier there was a single test case for all the sysl files.

springwiz:
> > I don’t see any parallelisation of unit tests. Probably better to keep that split out from this PR anyway. Please comment and update the description if necessary.
> 
> sysl2/sysl/exporter/swagger_exporter_test.go -> TestExportAll()
> Has multiple t.run() calls now. Earlier there was a single test case for all the sysl files.

Ok change removed, will create another PR with just the test case change. The description updated for this PR. Please have another look.

sysl2/sysl/cmd_export.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (p *exportCmd) Name() string            { return "export" }
+func (p *exportCmd) RequireSyslModule() bool { return true }
+
+func (p *exportCmd) Configure(app *kingpin.Application) *kingpin.CmdClause {
+	cmd := app.Command(p.Name(), "Export sysl to external types. Supported types: Swagger")
+	cmd.Flag("app-name",
+		"name of the sysl app defined in sysl model."+
+			" if there are multiple apps defined in sysl model,"+
+			" swagger will be generated only for the given app").Required().Short('a').StringVar(&p.appName)
```
anz-gordonj7: This flag isnt really required, if there are multiple apps defined and the flag is missing then flag an error, but otherwise there is little reason to force this flag on the command line.

sysl2/sysl/cmd_export.go
```diff
@@ -0,0 +1,58 @@
...
+	cmd.Flag("output", "output filename").Default("output.yaml").Short('o').StringVar(&p.out)
+	cmd.Flag("target", "export target").Default("swagger").Short('t').StringVar(&p.target)
+	cmd.Flag("format", "export format").Default("yaml").Short('f').StringVar(&p.format)
```
anz-gordonj7: These 3 flags don't make for a nice user interface. I'd say force the filename to be required (or maybe not, not a huge issue), remove the target flag completly, and change format to be exactly the same as the import flag options, default to "auto" (which means guess from the filename, .yaml == swagger yaml, .json ==swagger .json, etc)

sysl2/sysl/cmd_export.go
springwiz: So if you have 100's of apps in the Sysl file, you either export none or all? You don't give the user to choose the apps he wants to export. So no its not a great idea to drop the flag altogether. What we could do is to make the option "app-name" pick up the default value of "All".

anz-gordonj7: I didnt mean drop it fully, just remove the .Required() call and default to the default app name if not provided (or error out if more than 1 app is defined but the flag isnt provided

sysl2/sysl/cmd_export.go
springwiz: Done

sysl2/sysl/cmd_export.go
```diff
@@ -0,0 +1,103 @@
...
+			return err
+		}
+		output, err = swaggerExporter.SerializeOutput(p.format)
+		if err != nil {
+			return err
+		}
+	} else {
+		return fmt.Errorf("unsupported export format")
+	}
+	filename := fmt.Sprintf("%s_%s.%s", strings.Split(p.out, ".")[0], syslApp.Name.Part[0], p.format)
```
anz-gordonj7: This isnt safe for a few reasons:
1) "-o ../sometihng.json" will generate the filename "_appname.json" in the current dir instead of what is actually requested.
2) what if the app has multiple dots/parts?
3) p.format could be "swagger" which is a bad filename extension.


Solutions:
1) Maybe look at how sequence diagram "app_format" flag is implemented and do something similar for the filename, so it would allow a filename of "../out/%(appname)_generated.json" 
2) read the file extension with https://golang.org/pkg/path/filepath/#Ext

sysl2/sysl/cmd_export.go
```diff
@@ -0,0 +1,103 @@
...
+	cmd.Flag("output",
+		"export filename in the format filename.export_format, "+
+			" supported options: filename.json; filename.yaml").Required().Short('o').StringVar(&p.out)
+	EnsureFlagsNonEmpty(cmd)
+	return cmd
+}
+
+func (p *exportCmd) writeSwaggerForApp(syslApp *sysl.Application, logger *logrus.Logger) error {
+	var output []byte
+	var err error
```
anz-gordonj7: remove this var

sysl2/sysl/cmd_export.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		output, err = swaggerExporter.SerializeOutput(p.format)
+		if err != nil {
+			return err
+		}
+	} else {
+		return fmt.Errorf("unsupported export format")
+	}
+	filename := fmt.Sprintf("%s_%s.%s", strings.Split(p.out, ".")[0], syslApp.Name.Part[0], p.format)
+	fileWriteError := ioutil.WriteFile(filename, output, 0644)
```
anz-gordonj7: err := ioutil....

no need to keep adding new error vars unless you want to keep an old one for some reason

and you could just do `return ioutil.WriteFile()` in this case

sysl2/sysl/cmd_export.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	return nil
+}
+
+func (p *exportCmd) Execute(args ExecuteArgs) error {
+	err := p.determineOperationMode(p.out)
+	if err != nil {
+		return err
+	}
+	if p.appName == "all" {
```
anz-gordonj7: do the casefold compare so All and ALL are accepted

Also, this could be rewritten as 
```go
for _, app := range args.Module.GetApps() {
    switch p.appName {
         case app.Name.String(), "all":
                   err := p.writeSwaggerForApp(syslApp, args.Logger)
                  // maybe don't abort on an error, but post a Error logline instead
   }
}
```

springwiz: p.appName is defaulted to "all", upercasing to "ALL" and comparing doesn't help much.

sysl2/sysl/cmd_export.go
```diff
@@ -0,0 +1,103 @@
...
+func (p *exportCmd) Name() string            { return "export" }
+func (p *exportCmd) RequireSyslModule() bool { return true }
+
+func (p *exportCmd) Configure(app *kingpin.Application) *kingpin.CmdClause {
+	cmd := app.Command(p.Name(), "Export sysl to external types. Supported types: Swagger")
+	cmd.Flag("app-name",
+		"name of the sysl app defined in sysl model."+
+			" if there are multiple apps defined in sysl model,"+
+			" swagger will be generated only for the given app").Default("all").Short('a').StringVar(&p.appName)
+	cmd.Flag("output",
```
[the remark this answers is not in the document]
springwiz: I agree that the design is minimalistic but it works well. I agree that the user may not look at sysl --help but the hintoptions and error messages can provide the necessary info.

sysl2/sysl/cmd_export.go
```diff
@@ -0,0 +1,103 @@
...
+		"export filename in the format filename.export_format, "+
+			" supported options: filename.json; filename.yaml").Required().Short('o').StringVar(&p.out)
+	EnsureFlagsNonEmpty(cmd)
+	return cmd
+}
+
+func (p *exportCmd) writeSwaggerForApp(syslApp *sysl.Application, logger *logrus.Logger) error {
+	var output []byte
+	var err error
+	if p.mode == swaggerMode {
```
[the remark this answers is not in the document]
springwiz: Please look at the earlier comment.

sysl2/sysl/cmd_export.go
anz-gordonj7: OK, I need to work on ym review explanations, will try to explain myself more in the future.

I was showing you another way to write the if/else if to reduce the amount of repeated code (And I learnt that go allows non-static case values which is pretty cool)

sysl2/sysl/cmd_export.go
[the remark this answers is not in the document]
anz-gordonj7: It is very important to try to get user-facing changes as close to perfect as can be from the beginning, otherwise it adds a huge barrier to change later on. Even if the only people using the app are ourselves, and even if it is feature we may not use much, it is still important to get it right first go.

This is 100x more important with adding new API's because you never want to break other peoples code (the linux kernel has an almost concrete rule that no matter how big a bug or feature is, if it breaks external API or ABI it will be rejected).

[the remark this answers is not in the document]
springwiz: Design doesn't have to elaborate to work, it is important to get the things right I agree but that doesn't mean you should choose an overtly complex design each time.

sysl2/sysl/cmd_export.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	} else {
+		return fmt.Errorf("unsupported export format")
+	}
+	dir, err := filepath.Abs(filepath.Dir(filename))
+	if err != nil {
+		return err
+	}
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		logger.Debugf("Target folder exists %s", err)
```
anz-gordonj7: target exists or doesnt exist?

sysl2/sysl/cmd_export.go
```diff
@@ -0,0 +1,112 @@
...
+		err = os.Mkdir(dir, 0755)
```
anz-gordonj7: prefer one line version ```
if err := os.MkDir(); err != nil {
    return err
}```

Here and below (and generally)

sysl2/sysl/exporter/endpoint_exporter.go
```diff
@@ -130,41 +130,52 @@ func (e *EndpointExporter) setHTTPMethod(path string, endpoint *proto.Endpoint,
...
+	} else {
+		param = param.AsOptional()
+	}
+	if param.Type == object {
+		param.Schema.ExtraProps["$ref"] = "#/definitions/" + param.Format
+	}
+}
+
 func (e *EndpointExporter) setPathParams(endpoint *proto.Endpoint, op *spec.Operation) error {
+	var attrMap map[string]*proto.Attribute
```
anz-gordonj7: unnecessary, attrMap only used inside the loop

sysl2/sysl/exporter/endpoint_exporter.go
```diff
@@ -130,41 +130,52 @@ func (e *EndpointExporter) setHTTPMethod(path string, endpoint *proto.Endpoint,
...
 	for _, inParam := range endpoint.GetRestParams().GetUrlParam() {
+		attrMap = inParam.GetType().GetAttrs()
 		param := spec.PathParam(inParam.GetName())
```
anz-gordonj7: it looks like this line is now the only difference in these two loops, Can you pass `spec.PathParam`/`spec.QueryParam` as a func pointer to the setCommonAttributes() func to trim this down to just the last 2 lines?

sysl2/sysl/sysl_test.go
```diff
@@ -609,12 +609,73 @@ func TestMain2WithBinaryInfoCmd(t *testing.T) {
...
+func TestSwaggerAppExportCd(t *testing.T) {
 	t.Parallel()
 	logger, _ := test.NewNullLogger()
 	_, fs := testutil.WriteToMemOverlayFs(".")
-	main2([]string{"sysl", "export-swagger", "-o", "SIMPLE_SWAGGER_EXAMPLE.yaml", "-a", "testapp",
+	main2([]string{"sysl", "export", "-o", "SIMPLE_SWAGGER_EXAMPLE.yaml", "-a", "testapp",
 		"exporter/test-data/SIMPLE_SWAGGER_EXAMPLE.sysl"}, fs, logger, main3)
 	_, err := ioutil.ReadFile("SIMPLE_SWAGGER_EXAMPLE.yaml")
 	assert.NoError(t, err)
+	os.Remove("SIMPLE_SWAGGER_EXAMPLE.yaml")
```
anz-gordonj7: Please check the other tests that use WriteToMemOverlayFs. That fs implementation doesnt write anything to the real filesystem so this remove shouldnt be needed. I think you should be passing memfs to the main2() (where memfs is the _'ed return value)

sysl2/sysl/sysl_test.go
```diff
@@ -609,12 +609,73 @@ func TestMain2WithBinaryInfoCmd(t *testing.T) {
...
+}
+
+func TestSwaggerAppExportSd(t *testing.T) {
```
anz-gordonj7: ExportSd?

sysl2/sysl/sysl_test.go
```diff
@@ -609,12 +609,73 @@ func TestMain2WithBinaryInfoCmd(t *testing.T) {
...
+	t.Parallel()
+	logger, _ := test.NewNullLogger()
+	_, fs := testutil.WriteToMemOverlayFs(".")
+	main2([]string{"sysl", "export", "-o", "tests/SIMPLE_SWAGGER_EXAMPLE1.yaml", "-a", "testapp",
+		"exporter/test-data/SIMPLE_SWAGGER_EXAMPLE.sysl"}, fs, logger, main3)
+	_, err := ioutil.ReadFile("tests/SIMPLE_SWAGGER_EXAMPLE1.yaml")
+	assert.NoError(t, err)
+	os.Remove("tests/SIMPLE_SWAGGER_EXAMPLE1.yaml")
```
anz-gordonj7: as above (ditto the other tests in this file)

sysl2/sysl/cmd_export.go
```diff
@@ -0,0 +1,112 @@
...
+	return nil
+}
+
+func (p *exportCmd) Execute(args ExecuteArgs) error {
+	err := p.determineOperationMode(p.out)
+	if err != nil {
+		return err
+	}
+
+	var outputFileName string
```
anz-gordonj7: no need to create this out of the loop

sysl2/sysl/cmd_export.go
springwiz: Nope that doesn't allow you to log that there is actually a permission, it will simply create a directory if that doesn't exist. Besides the Mkdir will fail if the target folder exists.

sysl2/sysl/cmd_export.go
springwiz: ok

sysl2/sysl/exporter/endpoint_exporter.go
springwiz: for range loop in golang doesn't allow you to create and initialise new variables as part of loop initialisation as traditional for loop does. I don't want to create attrMap each time I loop. That is why attrMap has been created outside the loop.

sysl2/sysl/exporter/endpoint_exporter.go
springwiz: Obsession to DRY is bad and impacts the readability of the code.

sysl2/sysl/cmd_export.go
springwiz: Please see the comment on the attrMap earlier.

sysl2/sysl/sysl_test.go
springwiz: Was trying to test, whether the write succeeds in the target directory and not just the current directory, will name it to something better.

sysl2/sysl/cmd_export.go
anz-gordonj7:
```
if err := os.MkDir(); err != nil { 
   logger.Errorf("Error creating target folder; check permission")
    return err
}```

I dont understand your objection

springwiz: Got that

sysl2/sysl/sysl_test.go
springwiz: ok

sysl2/sysl/exporter/endpoint_exporter.go
springwiz: Was just trying to optimize, will change.

sysl2/sysl/exporter/endpoint_exporter.go
anz-gordonj7: 100% DISAGREE! when multiple sections/functions/whatever has almost the same code there is additional work needed by people coming back to the code, usually with questions like:
1. are the differences here what is actually required? Or is one buggy (and then which one?)
1. I have to fix a bug (or requirement change) in this code, how am I supposed to find out where else the same construct exists which needs the same change?

This link [pdf](https://people.cs.uchicago.edu/~shanlu/paper/TSE-CPMiner.pdf) isnt what I was looking for but its close enough, its a long article about why duplicated code sections are dangerous (from the point of view of how they were duplicated - copy + pasting code snippets), the link I was actually looking for was an article by (I think) Jon Carmack showing some of the copy+paste bugs which appeared in the ID software's game engines.

so No, DRY is absolutely important and is almost the #1 rule

springwiz: Ok may be my comment was a little awry, the disagree is in understanding what exactly is a repetition, it would be easier to discuss F2f.

PR #378, opened 2019-10-23
Feature/minor code fixes

anz-gordonj7:
* Add my local_ci.sh script which everyone should be using (or at least something similar) so we dont get golint-ci warnings in pull requests!

* Setup a generate.go file for the grammar output

@anz-bank/sysl-developers

### DIFF
--- a/sysl2/sysl/exporter/type_exporter.go
+++ b/sysl2/sysl/exporter/type_exporter.go
@@ -51,36 +51,31 @@ func (t *TypeExporter) populateTypes(syslTypes map[string]*proto.Type, swaggerTy
 			t.log.Warnf("Type matching failed %s", err)
 			return err
 		}
-		var ok bool
-		typeSchema.Format, ok = valueMap["format"]
-		typeSchema.Type = append(typeSchema.Type, valueMap["type"])
-		if !ok {
-			return fmt.Errorf("malformed sysl missing type info")
-		}
+		typeSchema.Format = valueMap.Format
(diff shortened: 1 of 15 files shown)


sysl2/sysl/exporter/type_exporter.go
```diff
@@ -51,63 +51,58 @@ func (t *TypeExporter) populateTypes(syslTypes map[string]*proto.Type, swaggerTy
...
-				attrValueMap, err = t.findSwaggerType(attV)
+				attrValueMap, err := t.findSwaggerType(attV)
 				if err != nil {
 					t.log.Warnf("Type matching failed %s", err)
 					return err
 				}
+				elementSchema.Format = attrValueMap.Format
+				elementSchema.Type = []string{attrValueMap.Type}
 			}
-			elementSchema.Format = attrValueMap["Format"]
```
anz-gordonj7: @springwiz this is why maps are bad...

springwiz: @anz-gordonj7 You mean missing type info in yml is due to maps being used?

anz-gordonj7: no, the key is incorrect. "**f**ormat" was used everywhere else, but this one used "**F**ormat"

springwiz: That is a "typo", I could have used constants. It still doesn't prove maps are bad.

springwiz: Thanks for fixing :-)

PR #379, opened 2019-10-24
Remove useless tests which occasionally fail because of timestamps

anz-gordonj7:
Remove tests which just check that the logger is hooked up to the error handler, not really useful and occasionally fails if the ttime changes as the test is running

@anz-bank/sysl-developers


anz-gordonj7:
> Don't have much clue on this functionality, but seems like valid test cases for SyslParserErrorListener. Why do we have to remove them?

@springwiz They are useless tests because the only thing they do is check that the callback writes something to the log and the correct timestamp. It was failed enough during my local builds that its driving me mad (the timestamp between the test happening and the log being written can be long enough to fail). Just want to remove them

sysl2/sysl/parse/error_listener_test.go
```diff
@@ -1,109 +0,0 @@
-package parse
```
anzdaddy: It would be better to loosen the check in `assertLog`.

nofun97:
the test caused #381 too

PR #383, opened 2019-10-28
Changes to the binary arg transform code:

anz-gordonj7:
* Remove the need to manually define callbacks for the NEQ op, simply check for a EQ and negate the result
* Add "in" support for maps (i.e '"foo" in foo.keys)
* Add unary negate support for binary values

@anz-bank/sysl-developers

### DIFF
--- a/sysl2/sysl/eval/binexprEval.go
+++ b/sysl2/sysl/eval/binexprEval.go
@@ -24,6 +24,9 @@ type Strategy interface {
 // DefaultBinExprStrategy is to evaluate lhs and rhs expr's first and then pass it to fn
 type DefaultBinExprStrategy struct{}
 
+// NegateBinExprStrategy is to evaluate the negative of the DefaultBinExprStrategy
+type NegateBinExprStrategy struct{}
+
 // LHSOverRHSStrategy binds rhs expression over each element of the LHS.
 // Assumes lhs is a collection
 type LHSOverRHSStrategy struct{}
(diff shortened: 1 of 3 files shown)


sysl2/sysl/eval/binexprEval.go
```diff
@@ -117,9 +115,26 @@ func (op DefaultBinExprStrategy) eval(txApp *sysl.Application, assign Scope, bin
...
 	}
+
 	panic(errors.Errorf("Unsupported operation:DefaultBinExprStrategy: %s", key))
 }
 
+func (op NegateBinExprStrategy) eval(txApp *sysl.Application, assign Scope, binexpr *sysl.Expr_BinExpr) *sysl.Value {
+	lhsValue := Eval(txApp, assign, binexpr.Lhs)
+	rhsValue := Eval(txApp, assign, binexpr.Rhs)
+
+	if binexpr.Op != sysl.Expr_BinExpr_NE {
```
gkanz: This should be the first thing checked.

sysl2/sysl/eval/binexprEval.go
```diff
@@ -117,9 +115,26 @@ func (op DefaultBinExprStrategy) eval(txApp *sysl.Application, assign Scope, bin
...
+		panic(errors.Errorf("Attempting to call Negate Startergy on a non == operator"))
+	}
+
+	key := makeKey(sysl.Expr_BinExpr_EQ, getValueType(lhsValue), getValueType(rhsValue))
```
gkanz: Why not  replace the remainder of this function with `return unaryNeg(DefaultBinExprStrategy{}.eval(txApp, assign, binexpr))`?

anz-gordonj7: Oh yeah, thats much cleaner :D

PR #386, opened 2019-10-29
added import checker

### LINKED ISSUE #375: Importing outside root
### Expected behaviour
Importing should either be relative to the sysl module or root. No '..' allowed

### Actual behaviour
It is possible to do something like `import '/../test'`. It is also possible to import outside root with the same example.

`import '/../test'`

nofun97:
Fixes #375 .

### Changes proposed in this pull request:
Importing is enforced. No longer be able to go above the root directory.

@anz-bank/sysl-developers


sysl2/sysl/syslutil/chroot_fs.go
```diff
@@ -70,3 +79,11 @@ func (fs *ChrootFs) Chmod(name string, mode os.FileMode) error {
 func (fs *ChrootFs) Chtimes(name string, atime time.Time, mtime time.Time) error {
 	return fs.fs.Chtimes(fs.join(name), atime, mtime)
 }
+
+func (fs *ChrootFs) openAllowed(name string) (err error) {
+	modulePath := filepath.Join(fs.root, name)
```
anz-gordonj7: will this work with the following dir structure: `/a/b/c/d/e` rooted at `/` module: `/a/b/c/d/e/foo.sysl` attempting to import `../../bar.sysl`?

I think you need to do a few calls to filepath.EvalSymLinks and filepath.Clean

sysl2/sysl/syslutil/chroot_fs.go
```diff
@@ -70,3 +79,11 @@ func (fs *ChrootFs) Chmod(name string, mode os.FileMode) error {
...
+	if !strings.HasPrefix(modulePath, fs.root) {
+		err = errors.New("import can not go up the directory")
```
anz-gordonj7: "permission denied, file outside root" would be a bettre error

nofun97: changed the error, also changed the way how it checks the import module, should be more foolproof.

sysl2/sysl/syslutil/chroot_fs.go
nofun97: a lot of filepath functions already call `Clean` automatically, but will add the EvalSymLinks in next commit

sysl2/sysl/syslutil/chroot_fs.go
```diff
@@ -70,3 +78,10 @@ func (fs *ChrootFs) Chmod(name string, mode os.FileMode) error {
 func (fs *ChrootFs) Chtimes(name string, atime time.Time, mtime time.Time) error {
 	return fs.fs.Chtimes(fs.join(name), atime, mtime)
 }
+
+func (fs *ChrootFs) openAllowed(name string) (err error) {
+	if strings.Contains(name, "..") {
+		return errors.New("permission denied, file outside root")
+	}
+	return
+}
```
anzdaddy: `abc/../..def` is valid.

anz-gordonj7: If filepath.Clean() is used the path will never have .. in it anywah

anzdaddy: Only inner refs such as `abc/..` will be collapsed. Prefix `../` will remain, so the solution is to apply `filepath.Clean` and then error out if `name == ".." || strings.HasPrefix(name, "../)`.

nofun97:
WIP: currently adding more tests on ChrootFs

sysl2/sysl/syslutil/chroot_fs.go
```diff
@@ -1,8 +1,11 @@
 package syslutil
 
 import (
+	"errors"
+	"log"
```
golangcibot: "log" imported but not used (from `typecheck`)

sysl2/sysl/syslutil/chroot_fs.go
nofun97: resolved in the test

nofun97:
closed, replaced with #399

PR #398, opened 2019-11-11
Create new aliasing syntax

### LINKED ISSUE #229: More concise aliases
Before:
```
    !alias FooBar:
        string
```

After:
```
    !alias FooBar = string
```

Joshcarp:
Fixes #229.

- Add additional alias syntax to support 

```
!alias Foo : string
```

This acts the same way as 
```
    !alias FooBar:
        string
```
Including annotations and modifiers

(Note `alias Foo = string` is a much more difficult task because `TEXT_LINE` matches with `=` so quite a lot of rules need to be changes to accomodate this change)

@anz-bank/sysl-developers


sysl2/sysl/parse/tests/alias_new.sysl
```diff
@@ -0,0 +1,2 @@
+Model:
```
anzdaddy: Avoid temporal elements in filenames. It won't be new forever. Suggest `alias_inline.sysl`.

sysl2/sysl/grammar/SyslParser.g4
```diff
@@ -384,8 +384,10 @@ view
     (ARROW_RIGHT view_return_type)?
     ( attribs_or_modifiers? COLON expr_block | abstract_view {$abstractView=true;} );
 
-alias: ALIAS name_str attribs_or_modifiers? COLON
-        INDENT annotation* (types | collection_type) DEDENT;
+alias: ALIAS name_str attribs_or_modifiers? COLON (
+        (annotation* (types | collection_type))
+        | (INDENT annotation* (types | collection_type) DEDENT)
```
anz-gordonj7: these two lines should be combined into another rule to remove duplication:
```
alias_def: annotation* (types | collection_type)
alias: ALIAS name_str attribs_or_modifiers? COLON (alias_def | INDENT alias_def DEDENT)
```

Otherwise this PR looks fine

sysl2/sysl/parse/Parser_test.go
```diff
@@ -496,6 +496,12 @@ func TestCrash(t *testing.T) {
 	testParseAgainstGolden(t, "tests/crash.sysl", "")
 }
 
+func TestAlias(t *testing.T) {
+	t.Parallel()
+
+	testParseAgainstGolden(t, "tests/alias_new.sysl", "")
```
anz-gordonj7: I really dislike this style of tests!

anzdaddy: What exactly?

sysl2/sysl/grammar/SyslParser.g4
Joshcarp: need a way to access `ctx.alias_def.annotation()` in the listener in order to achieve this

PR #400, opened 2019-11-12
Add single unary expression

gkanz:
- The "single" unary expression will be added.

@anz-bank/sysl-developers


anzdaddy:
Does this support parsing the operator?

gkanz:
> Does this support parsing the operator?

That was already there, it was throwing an exception when it was used. Here it is in the lexer: https://github.com/anz-bank/sysl/blob/1222cec1e91e3ed2dff102bb384870fc4b99224a/sysl2/sysl/grammar/SyslLexer.g4#L314.

PR #403, opened 2019-11-12
Allow trace level logging to be enabled

gkanz:
- Very vervose trace logging can be enabled via the command line

@anz-bank/sysl-developers


anzdaddy:
This appears to be a duplicate of #401.

PR #414, opened 2019-11-14
Root flag is now optional

nofun97:
- Root flag in the CLI is now optional, sysl can find the project root itself if root flag is not defined. ChrootFs will be rooted to this root.
- Root can be defined either by adding the path `--root` in the CLI or using a root marker which marks the location of the project root (currently it is a directory called `.sysl`). Root marker have to be either in the same directory or any parent directory of the module argument in the command. Root flag takes priority and will be used if it is defined.
- If root is not found, the directory of the module argument is set as the root.

@anz-bank/sysl-developers

### DIFF
--- a/sysl2/sysl/cmd_runner.go
+++ b/sysl2/sysl/cmd_runner.go
@@ -123,12 +123,18 @@ func EnsureFlagsNonEmpty(cmd *kingpin.CmdClause, excludes ...string) {
 }
 
 func (r *cmdRunner) getProjectRoot(fs afero.Fs, logger *logrus.Logger) error {
-	syslRootPath, err := findRootFromSyslModule(r.module, fs)
+	rootIsDefined := r.Root != ""
+
+	modulePath := r.module
+	if rootIsDefined {
+		modulePath = filepath.Join(r.Root, r.module)
+	}
+
+	syslRootPath, err := findRootFromSyslModule(modulePath, fs)
 	if err != nil {
 		return err
 	}
 
-	rootIsDefined := r.Root != ""
 	rootMarkerExists := syslRootPath != ""
 
 	switch {
--- a/sysl2/sysl/cmd_runner.go
+++ b/sysl2/sysl/cmd_runner.go
@@ -33,8 +33,7 @@ func (r *cmdRunner) Run(which string, fs afero.Fs, logger *logrus.Logger) error 
 			var mod *sysl.Module
 			var err error
 			if cmd.RequireSyslModule() {
-				err = r.getProjectRoot(fs, logger)
-				if err != nil {
+				if err = r.setProjectRoot(fs, logger); err != nil {
 					return err
 				}
 
@@ -51,8 +50,6 @@ func (r *cmdRunner) Run(which string, fs afero.Fs, logger *logrus.Logger) error 
 }
 
 func (r *cmdRunner) Configure(app *kingpin.Application) error {
-	app.UsageTemplate(kingpin.SeparateOptionalFlagsUsageTemplate)
-
 	commands := []Command{
 		&protobuf{},
 		&intsCmd{},
@@ -122,7 +119,7 @@ func EnsureFlagsNonEmpty(cmd *kingpin.CmdClause, excludes ...string) {
 	cmd.PreAction(fn)
 }
 
-func (r *cmdRunner) getProjectRoot(fs afero.Fs, logger *logrus.Logger) error {
+func (r *cmdRunner) setProjectRoot(fs afero.Fs, logger *logrus.Logger) error {
 	rootIsDefined := r.Root != ""
 
 	modulePath := r.module
@@ -137,27 +134,30 @@ func (r *cmdRunner) getProjectRoot(fs afero.Fs, logger *logrus.Logger) error {
 
 	rootMarkerExists := syslRootPath != ""
 
-	switch {
-	case rootIsDefined:
+	if rootIsDefined {
 		if rootMarkerExists {
 			logger.Warningf(rootMarkerExistsWarning, syslRootMarker, syslRootPath, r.Root)
 		} else {
 			logger.Warningf(noRootMarkerWarning, syslRootMarker, r.Root)
 		}
-	case !rootIsDefined && !rootMarkerExists:
-		// uses the module directory as the root, changing the module to be relative to the root
-		r.Root = filepath.Dir(r.module)
-		r.module = filepath.Base(r.module)
-		logger.Warningf(noRootWarning, syslRootMarker, r.Root)
-	case !rootIsDefined && rootMarkerExists:
-		r.Root = syslRootPath
-		absModulePath, err := filepath.Abs(r.module)
-		if err != nil {
-			return err
-		}
-		r.module, err = filepath.Rel(r.Root, absModulePath)
-		if err != nil {
-			return err
+	} else {
+		if rootMarkerExists {
+			r.Root = syslRootPath
+
+			// module has to be relative to the root
+			absModulePath, err := filepath.Abs(r.module)
+			if err != nil {
+				return err
+			}
+			r.module, err = filepath.Rel(r.Root, absModulePath)
+			if err != nil {
+				return err
+			}
+		} else {
+			// uses the module directory as the root, changing the module to be relative to the root
+			r.Root = filepath.Dir(r.module)
+			r.module = filepath.Base(r.module)
+			logger.Warningf(noRootWarning, syslRootMarker, r.Root)
 		}
 	}
 
@@ -165,7 +165,6 @@ func (r *cmdRunner) getProjectRoot(fs afero.Fs, logger *logrus.Logger) error {
 }
 
 func findRootFromSyslModule(modulePath string, fs afero.Fs) (string, error) {
-	// Takes the closest root marker
 	currentPath, err := filepath.Abs(modulePath)
 	if err != nil {
 		return "", err
@@ -176,8 +175,8 @@ func findRootFromSyslModule(modulePath string, fs afero.Fs) (string, error) {
 		return "", err
 	}
 
+	// Keep walking up the directories to find nearest root marker
 	for {
-		// Keep walking up the directories
 		currentPath = filepath.Dir(currentPath)
 		exists, err := afero.Exists(fs, filepath.Join(currentPath, syslRootMarker))
 		reachedRoot := currentPath == systemRoot || (err != nil && os.IsPermission(err))
--- a/sysl2/sysl/cmd_runner.go
+++ b/sysl2/sysl/cmd_runner.go
@@ -14,10 +14,7 @@ import (
 )
 
 const (
-	syslRootMarker          = ".sysl"
-	noRootMarkerWarning     = "%s is not defined but root flag is defined in %s"
-	rootMarkerExistsWarning = "%s found in %s but will use %s instead"
-	noRootWarning           = "root and %s are undefined, %s will be used instead"
+	syslRootMarker = ".sysl"
 )
 
 type cmdRunner struct {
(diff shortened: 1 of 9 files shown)


nofun97:
Another feature will be added which restricts importing from the root if root is not defined will be added as another PR to keep this PR small. @anzdaddy

sysl2/sysl/cmd_runner.go
```diff
@@ -11,6 +13,13 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
+const (
+	syslRootMarker          = ".sysl"
+	noRootMarkerWarning     = "%s is not defined but root flag is defined in %s"
+	rootMarkerExistsWarning = "%s found in %s but will use %s instead"
+	noRootWarning           = "root and %s are undefined, %s will be used instead"
```
anz-gordonj7: Move these down to where they are actually used, with them up here someone looking back at the warning code needs to scroll up and down to see what will be printed in what order. It also makes it more annoying to grep the source for the text (i.e you're running the program, see that text in the console and want to find where it comes from, its now a 2 step to get there).

There is also no value in making them const instead of just using them directly

sysl2/sysl/cmd_runner.go
```diff
@@ -24,6 +33,11 @@ func (r *cmdRunner) Run(which string, fs afero.Fs, logger *logrus.Logger) error
 			var mod *sysl.Module
 			var err error
 			if cmd.RequireSyslModule() {
+				err = r.getProjectRoot(fs, logger)
```
anz-gordonj7: nitpick: when a func only returns an error, put it in a single line ```if err := func(); 
err != nil { 
return err 
}```

sysl2/sysl/cmd_runner.go
```diff
@@ -37,6 +51,8 @@ func (r *cmdRunner) Run(which string, fs afero.Fs, logger *logrus.Logger) error
 }
 
 func (r *cmdRunner) Configure(app *kingpin.Application) error {
+	app.UsageTemplate(kingpin.SeparateOptionalFlagsUsageTemplate)
```
anz-gordonj7: this line should be when the app is created

sysl2/sysl/cmd_runner.go
```diff
@@ -105,3 +121,73 @@ func EnsureFlagsNonEmpty(cmd *kingpin.CmdClause, excludes ...string) {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func findRootFromSyslModule(modulePath string, fs afero.Fs) (string, error) {
+	// Takes the closest root marker
```
anz-gordonj7: what is this comment for?

sysl2/sysl/cmd_runner.go
```diff
@@ -105,3 +121,73 @@ func EnsureFlagsNonEmpty(cmd *kingpin.CmdClause, excludes ...string) {
 
 	cmd.PreAction(fn)
 }
+
+func (r *cmdRunner) getProjectRoot(fs afero.Fs, logger *logrus.Logger) error {
```
anz-gordonj7: function naming is extremely important for code readability and maintainability. `get*()` functions are expected to return the thing it is getting. I'm struggling to figure out what this function actually does.

sysl2/sysl/cmd_runner.go
```diff
@@ -105,3 +121,73 @@ func EnsureFlagsNonEmpty(cmd *kingpin.CmdClause, excludes ...string) {
...
+	rootIsDefined := r.Root != ""
+
+	modulePath := r.module
+	if rootIsDefined {
+		modulePath = filepath.Join(r.Root, r.module)
+	}
+
+	syslRootPath, err := findRootFromSyslModule(modulePath, fs)
```
anz-gordonj7: I'm not sure I see the point of searching for a .sysl/ if the root arg was supplied, just unconditionally warn that the are is no longer required and should be replaced with the folder marker. Doing that would simplify this func

sysl2/sysl/cmd_runner.go
```diff
@@ -105,3 +121,73 @@ func EnsureFlagsNonEmpty(cmd *kingpin.CmdClause, excludes ...string) {
...
+	rootMarkerExists := syslRootPath != ""
+
+	switch {
+	case rootIsDefined:
+		if rootMarkerExists {
+			logger.Warningf(rootMarkerExistsWarning, syslRootMarker, syslRootPath, r.Root)
+		} else {
+			logger.Warningf(noRootMarkerWarning, syslRootMarker, r.Root)
+		}
+	case !rootIsDefined && !rootMarkerExists:
```
anz-gordonj7: PLEASE be consistent with your codee style. You used a if inside the case above but this uses extra bool e
expressions. Should have just made it:
```
if rootIsDefined {
   if roootMarkerExists { } else {}
} else {
   if roootMarkerExists { } else {}
}

```

sysl2/sysl/cmd_runner_test.go
```diff
@@ -33,3 +59,186 @@ func TestEnsureFlagsNonEmpty(t *testing.T) {
...
+			expectedRoot:     filepath.Dir(undefinedRoot.files[0]),
+			expectedLog:      noRootLog,
+			rootMarkerExists: false,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(ts folderTestStructure) func(t *testing.T) {
+			return func(tt *testing.T) {
+				tt.Parallel()
```
anz-gordonj7: this could just be
```
t.Run(test.name, func(t *testing.T) {
    t.Parallel()
    ....
})  
```

sysl2/sysl/cmd_runner.go
nofun97: I'm using them in the test to check for logs. Should I copy the string to the test instead?

sysl2/sysl/cmd_runner_test.go
nofun97: `scopelint` would give out warnings if I do that

anz-gordonj7: should be fine, thats being done in other tests in sysl

sysl2/sysl/cmd_runner.go
anz-gordonj7: im not a fan of testing for log messages TBH

sysl2/sysl/cmd_runner.go
nofun97: It just warns the user that a `.sysl` is found, but if user still wants to set their own root that is still allowed. Would that warning be useful or unnecessary?

sysl2/sysl/cmd_runner.go
nofun97: changed to that

sysl2/sysl/cmd_runner.go
nofun97: moved to `sysl.go` in `main3()`

sysl2/sysl/cmd_runner.go
nofun97: Part of old code, it's removed

sysl2/sysl/cmd_runner.go
nofun97: changed to `setProjectRoot`

sysl2/sysl/cmd_runner.go
nofun97: fixed

sysl2/sysl/cmd_runner_test.go
```diff
@@ -33,3 +59,184 @@ func TestEnsureFlagsNonEmpty(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			logger, hook := log.NewNullLogger()
+			fs := afero.NewMemMapFs()
+			syslutil.BuildFolderTest(t, fs, test.structure.folders, test.structure.files)
```
golangcibot: Using the variable on range scope `test` in function literal (from `scopelint`)

sysl2/sysl/cmd_runner_test.go
```diff
@@ -33,3 +59,184 @@ func TestEnsureFlagsNonEmpty(t *testing.T) {
...
+
+			r := &cmdRunner{Root: test.root, module: test.module}
```
golangcibot: Using the variable on range scope `test` in function literal (from `scopelint`)

sysl2/sysl/cmd_runner_test.go
```diff
@@ -33,3 +59,184 @@ func TestEnsureFlagsNonEmpty(t *testing.T) {
...
+
+			require.NoError(t, r.setProjectRoot(fs, logger))
+			require.Equal(t, test.expectedRoot, r.Root)
```
golangcibot: Using the variable on range scope `test` in function literal (from `scopelint`)

sysl2/sysl/cmd_runner_test.go
nofun97: got the warnings, ignore it?

sysl2/sysl/cmd_runner.go
nofun97: okay, removed log testings and constants, moved the warnings to where it is used

sysl2/sysl/cmd_runner_test.go
```diff
@@ -33,3 +59,184 @@ func TestEnsureFlagsNonEmpty(t *testing.T) {
...
+			module:           undefinedRoot.files[0],
+			structure:        undefinedRoot,
+			expectedRoot:     filepath.Dir(undefinedRoot.files[0]),
+			expectedLog:      noRootLog,
+			rootMarkerExists: false,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
```
anz-gordonj7: add `test := test` before the run lne to get rid of the lint errors

nofun97: fixed

sysl2/sysl/cmd_runner.go
nofun97: @anzdaddy

anzdaddy: This was my suggestion, but I'm not particularly wedded to one approach or another. We'll learn and adapt.

sysl2/sysl/cmd_runner.go
```diff
@@ -11,6 +13,10 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
+const (
+	syslRootMarker = ".sysl"
+)
```
anzdaddy: nit: simplify
```go
const syslRootMarker = ".sysl"
```

sysl2/sysl/cmd_runner.go
```diff
@@ -51,8 +61,8 @@ func (r *cmdRunner) Configure(app *kingpin.Application) error {
 	r.commands = map[string]Command{}
 
 	app.Flag("root",
-		"sysl root directory for input model file (default: .)").
-		Default(".").StringVar(&r.Root)
+		"sysl root directory for input model file. If root is not found, the module directory becomes the root").
```
anzdaddy: This docstring implies that, in the absence of a root, `import /foo` will look for a file in the same directory as the importing file. It should fail.

sysl2/sysl/cmd_runner_test.go
```diff
@@ -1,13 +1,32 @@
...
+	"github.com/stretchr/testify/require"
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
+const (
+	currentWorkingDirectory = "."
+)
+
+type folderTestStructure struct {
+	name, module, root, expectedRoot, rootMarkerPath string
```
anzdaddy: This is getting a bit long. One line per field might be better in this case.

PR #422, opened 2019-11-17
Added absolute import restrictions

### LINKED ISSUE #421: Absolute import restriction
### Expected behaviour
Absolute path import or importing from root in sysl module should fail if root is not found (root flag not defined and `.sysl` not found)
```
import /path/to/module
```

nofun97:
Fixes #421 .

Importing from root (or absolute path import) will return an error if root is not found (root flag is not set and `.sysl` not found).

@anz-bank/sysl-developers

### DIFF
--- a/sysl2/sysl/eval/unaryEval.go
+++ b/sysl2/sysl/eval/unaryEval.go
@@ -9,7 +9,8 @@ type unaryFunc func(*sysl.Value) *sysl.Value
 
 //nolint:gochecknoglobals
 var unaryFunctions = map[sysl.Expr_UnExpr_Op]unaryFunc{
-	sysl.Expr_UnExpr_NEG: unaryNeg,
+	sysl.Expr_UnExpr_NEG:    unaryNeg,
+	sysl.Expr_UnExpr_SINGLE: unarySingle,
 }
 
 func evalUnaryFunc(op sysl.Expr_UnExpr_Op, arg *sysl.Value) *sysl.Value {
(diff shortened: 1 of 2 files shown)


sysl2/sysl/parse/parse.go
```diff
@@ -2,6 +2,7 @@ package parse
 
 import (
 	"fmt"
+	"log"
```
golangcibot: "log" imported but not used (from `typecheck`)

sysl2/sysl/parse/parse.go
```diff
@@ -127,8 +129,12 @@ func (p *Parser) Parse(filename string, fs afero.Fs) (*sysl.Module, error) {
 
 		for len(listener.imports) > 0 {
 			source = listener.imports[0]
+
 			listener.imports = listener.imports[1:]
 			if _, has := imported[source.filename]; !has {
+				if !p.rootFound && strings.HasPrefix(source.filename, "/") {
+					return nil, Exitf(AbsImportError, "error importing: import from root is not allowed if root is undefined")
```
anz-gordonj7: Didn't we talk about leaving this functionality to the filesystem? If root is found the chroot will forbid the import, if it isn't then the chroot will use the real / as root and allow. The parser should not be modified

nofun97: If this is implemented in chrootFs, that means chrootFs should be aware when it starts importing (so first module load should be ignored) and I thought that would be more complex since chrootFs is used a lot. Putting it in `Parse` seems to be simpler.

sysl2/sysl/parse/Parser_test.go
```diff
@@ -115,7 +115,12 @@ func parseComparable(
 	filename, root string,
 	stripSourceContext bool,
 ) (*sysl.Module, error) {
-	module, err := NewParser().Parse(filename, syslutil.NewChrootFs(afero.NewOsFs(), root))
+	// realRoot := root
```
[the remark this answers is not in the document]
nofun97: removed

sysl2/sysl/cmd_runner.go
```diff
@@ -124,11 +125,12 @@ func (r *cmdRunner) setProjectRoot(fs afero.Fs, logger *logrus.Logger) error {
 
 	syslRootPath, err := findRootFromSyslModule(modulePath, fs)
 	if err != nil {
-		return err
+		return false, err
 	}
 
 	rootMarkerExists := syslRootPath != ""
 
+	rootFound := true
```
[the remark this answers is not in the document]
nofun97: fixed

PR #426, opened 2019-11-19
Auto-Generate sysl for codegen grammar.

springwiz:
Fixes # .

- First step to allowing grammar to double up as validation schema.
- Extracts sysl from codegen grammar.
- Next step would be to use the same as part of codegen removing the need for sysl.

@anz-bank/sysl-developers


sysl2/sysl/importer/grammar.go
```diff
@@ -0,0 +1,205 @@
...
+	info := SyslInfo{
+		OutputData:  outputData,
+		Title:       g.grammarParam.startRule,
+		Description: outputData.AppName + " Grammar",
+	}
+
+	startRule := g.grammar.GetRules()[g.grammarParam.startRule]
+	ruleType := g.createRule(startRule, g.grammarParam.startRule)
+	g.types.types = append([]Type{ruleType}, g.types.types...)
+	g.visitedMap[g.grammarParam.startRule] = true
```
anz-gordonj7: I dont understand the need to specify the start rule,  The sysl file should contain every rule found in the grammar, not just the ones accessable from the start rule.

sysl2/sysl/importer/grammar.go
```diff
@@ -0,0 +1,205 @@
...
+	case `goFile`:
+		outputData = OutputData{
+			AppName: "Go",
+		}
+	case `javaFile`:
+		outputData = OutputData{
+			AppName: "Java",
+		}
+	default:
+		return "", fmt.Errorf("unexpected entry rule")
```
anz-gordonj7: The whole point of this feature is so new grammars don't need sysl files, why are you limiting support to go and java?!!?!?

sysl2/sysl/importer/grammar.go
springwiz: Well according to my understanding the AST for the chosen grammar would have only 1 root, and that is the start rule. We could deduce the start rule automatically rather than the user passing it over the command line.

springwiz: Anyways the intent is to use it more from codegen rather than importer.

anz-gordonj7: It wanting to be consumed by the codeegen code, and being able to geneerate sysl from grammar are two related but not the same thing. The sysl generation doesnt need to know the start rule.

springwiz: Taken care

sysl2/sysl/importer/grammar.go
springwiz: ok

sysl2/sysl/cmd_import.go
```diff
@@ -61,14 +63,29 @@ func (p *importCmd) Execute(args ExecuteArgs) error {
 	case modeXSD:
 		args.Logger.Infof("Using xsd importer\n")
 		imp = importer.LoadXSDText
+	case modeGrammar:
+		args.Logger.Infof("Using grammar importer\n")
+		p.OutputData.AppName, err = findGrammarType(p.filename)
```
anz-gordonj7: Why are you not simply using the app-name flag that is passed in on the command line? (and revert the change to make it optional)

sysl2/sysl/importer/grammar.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	switch quant.Union.(type) {
+	case *ebnfGrammar.Quantifier_Optional:
+		fieldType = fmt.Sprintf("%s?", fieldType)
+	case *ebnfGrammar.Quantifier_ZeroPlus:
+		fieldType = fmt.Sprintf("sequence of %s?", fieldType)
+	case *ebnfGrammar.Quantifier_OnePlus:
+		fieldName = fmt.Sprintf("%s(1..)", fieldName)
+	default:
+		g.logger.Warnf("unexpected quantifier type")
```
anz-gordonj7: This should probably be a fatal error, not a warning

sysl2/sysl/importer/grammar.go
```diff
@@ -0,0 +1,190 @@
...
+		g.visitedMap[name] = true
+	}
+	for _, seq := range rule.Choices.Sequence {
+		for _, term := range seq.Term {
+			switch x := term.Atom.Union.(type) {
+			case *ebnfGrammar.Atom_String_:
+				// hit the leaf node
+				// return
+			case *ebnfGrammar.Atom_Choices:
+				choiceField := g.quantifyUnionType("__Choice_"+name, "__Choice_"+name, term.Quantifier)
```
anz-gordonj7: `"__Choice_"+name` Used multiple times in this switch, make it a local var

sysl2/sysl/importer/grammar_test.go
```diff
@@ -0,0 +1,44 @@
+package importer
```
anz-gordonj7: There is already a standard test format for this package, use that by just adding another test func to the end of importer_tests.go instead of duplicating the test code.

sysl2/sysl/importer/tests-grammar/go.gen.g
```diff
@@ -0,0 +1,57 @@
+goFile: PackageClause Comment? '\n' ImportDecl? '\n' TopLevelDecl+ '\n';
```
anz-gordonj7: This is not a useful "test". Tests need to test 1 thing and 1 thing only. Using this test input, when a test fails the developer will have to then go digging "Something broke" is not helpful. "Union types are broken" Is much more useful outcomes.

Remove this and the java one, add multiple TINY test input files to get the required 99% coverage.

At a minimum I'd expect the following tests:
1. Just a single rule with only a single string
1. 2 rules, one referring to the other
1. simple but with withe * and ? qualifiers
1. simple with the choices (and choices with qualifiers).

sysl2/sysl/importer/grammar.go
```diff
@@ -0,0 +1,190 @@
...
+
+				// visit the unvisited node
+				if !g.visitedMap[x.Rulename.GetName()] {
+					if rule, exists := g.grammar.Rules[x.Rulename.GetName()]; exists {
+						childRuleType = g.createRule(g.grammar.Rules[rule.Name.Name], rule.Name.Name)
+						g.types.types = append([]Type{childRuleType}, g.types.types...)
+					}
+				}
+			default:
+				g.logger.Warnf("unexpected term type")
```
anz-gordonj7: fatal, not warn

sysl2/sysl/importer/grammar.go
```diff
@@ -0,0 +1,190 @@
...
+		OutputData: OutputData{
+			AppName: g.grammarName,
+		},
+		Title:       g.grammarName + "File",
+		Description: g.grammarName + " Grammar",
+	}
+	for keyRuleName, keyRule := range g.grammar.GetRules() {
+		if _, visited := g.visitedMap[keyRuleName]; !visited {
+			ruleType := g.createRule(keyRule, keyRuleName)
+			g.types.types = append([]Type{ruleType}, g.types.types...)
```
anz-gordonj7: `g.types.types = append(g.types.types, ruleType)`

sysl2/sysl/cmd_import.go
springwiz: The app-name is already available as the grammar name, why not use that?

sysl2/sysl/importer/grammar.go
springwiz: It already caters to all the know types as of now, tomorrow if the protobuff changes, I don't want the conversion to fail but let it proceed with a warning.

sysl2/sysl/importer/grammar.go
springwiz: Refer to earlier comment.

sysl2/sysl/importer/tests-grammar/go.gen.g
springwiz: I am not sure if the amount of effort involved justifies the payout.

anz-gordonj7: Please stop with this attitude! If these tests plus whatever else are needed to get to complete coverage takes more than 2 hours it almost certainly means the implementation is incredibly fragile. Test-after is extremely poor engineering. Test-driven design should be part of your standard work flow.

sysl2/sysl/importer/grammar.go
anz-gordonj7: tomorrow if the protobuf changes we 100% want to make sure the person who changes it fixes the importer to deal with it. Make this fatal

sysl2/sysl/importer/grammar.go
anz-gordonj7: ditto

sysl2/sysl/importer/tests-grammar/go.gen.g
anz-gordonj7: Apologies for replying before my morning coffee, after looking back I realised my remark was far more aggressive than necessary.

I don't think there needs to be more than half a dozen tests to get full coverage for this code, and each test need only be a few lines of grammar. I think the effort then is very low regardless of the possible payout

springwiz: We developers are folks passionate about technology and solid code. so no worries mate!  I have parked this PR for now and starting working on gRPC, will fix this and others once I am through with that.

PR #427, opened 2019-11-19
Issue/407/unmarshal correct type

### LINKED ISSUE #407: Need validation before unmarshall response in restlib
When we generate a client file in golang with two expected kinds of response it will generate code like `responses := []interface{}{&A{}, &B{}}`
This response will pass into `restlib.DoHTTPRequest()`
In this method's

```go
for _, respInterface := range response {
	err = json.Unmarshal(respBody, respInterface)
	if err == nil {
		return makeHTTPResult(httpResponse, respBody, respInterface), nil
	}
} 
```

It didn't do the validation before `unmarshall`, so this `unmarshall` method never failed, it always marshall the `respBody` to the first `respInterface` come up from the for loop.
In this case, the response body will always return as &A{}.

gkanz:
Fixes #407 (partially, see below)

The changes here are used by the SYSL->Go transforms to unmarshal the correct type. The inherent assumption is that there will be a 0..1 return types for OK conditions and 0..1 return types for error conditions.

An internal version of restlib was also updated to make use of this new functionality, this should be folded back in to the public restlib. However, that is outside the scope of this repository.

@anz-bank/sysl-developers

### DIFF
--- a/src/sysl/importers/import_openapi.py
+++ b/src/sysl/importers/import_openapi.py
@@ -173,15 +173,23 @@ class OpenApiTranslator:
                             schema = schema['schema']
 
                 if schema is not None:
-                    if schema.get('type') == 'array':
-                        returnValues['sequence of ' + schema['items']['$ref'].rpartition('/')[2]] = True
-                    else:
-                        returnValues[schema['$ref'].rpartition('/')[2]] = True
+                    varName = ""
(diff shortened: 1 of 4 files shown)


gkanz:
Agree, I'll create an issue. I would rather not have to use Python at all.

Issue #397, opened 2019-11-11
Use context more


anzdaddy:
### Expected behaviour

API use context for cross-cutting parameters:

```go
func f(ctx context.Context, arg1, arg2, arg3 something)
```

### Actual behaviour

Cross-cutting parameters pollute APIs.

```go
func f(arg1, arg2, arg3 something, fs afero.Fs, logger *logrus.Logger)
```

anz-gordonj7:
How do we decide what is appropriate for context and what isnt?

nofun97:
So far, logger and file systems (possibly) are going to be put in context since they're used almost everywhere.

anzdaddy:
There's no hard line. It'll always come down to making a judgment call. Any API that behaves non-deterministically likely belongs in context. To illustrate, `import "math/rand" … rand.Int()` should become something like `import "…/contextual/rand" … rand.Int(ctx)` whereas `rand.New(rand.NewSource(123))` can stay as is.

PR #428, opened 2019-11-19
first draft of the syslLogger library

nofun97:
Part of #397 .

This is the first draft of the sysl logger design. It is a work in progress.

@anz-bank/sysl-developers

### DIFF
--- a/docs/ideas/contextual-logging-example/main.go
+++ b/docs/ideas/contextual-logging-example/main.go
@@ -2,7 +2,8 @@ package main
 
 import (
 	"context"
-	"github.com/anz-bank/sysl/docs/ideas/contextual-logging-example/syslLogger"
+
+	logger "github.com/anz-bank/sysl/docs/ideas/contextual-logging-example/syslLogger"
 )
 
 func main() {
@@ -31,7 +32,7 @@ func main() {
 	 * Expected to log
 	 * 2006-01-02T15:04:05Z07:00 id=12345 DEBUG This is a debug with context specific field again (main.go:34)
 	 */
-	 logger.Debug(ctx, "This is a debug with context specific field again")
+	logger.Debug(ctx, "This is a debug with context specific field again")
 }
 
 func setupLogger(ctx context.Context) context.Context {
--- a/docs/ideas/contextual-logging-example/main.go
+++ b/docs/ideas/contextual-logging-example/main.go
@@ -1,0 +1,40 @@
+package main
+
+import (
+	"context"
+	"github.com/anz-bank/sysl/docs/ideas/contextual-logging-example/syslLogger"
+)
+
+func main() {
+	ctx := context.Background()
+	ctx = setupLogger(ctx)
+
+	ctx = logger.AddFields(ctx, map[string]interface{}{
+		"id": "12345",
+	})
+
+	/**
+	 * Expected to log
+	 * 2006-01-02T15:04:05Z07:00 id=12345 DEBUG This is a debug with context specific field (main.go:18)
+	 */
+	logger.Debug(ctx, "This is a debug with context specific field")
+
+	/**
+	 * Expected to log
+	 * 2006-01-02T15:04:05Z07:00 id=12345 log=hello DEBUG This is a debug with log specific field (main.go:26)
+	 */
+	logger.LogFields(ctx, map[string]interface{}{
+		"log": "hello",
+	}).Debug("This is a debug with log specific field")
+
+	/**
+	 * Expected to log
+	 * 2006-01-02T15:04:05Z07:00 id=12345 DEBUG This is a debug with context specific field again (main.go:34)
+	 */
+	 logger.Debug(ctx, "This is a debug with context specific field again")
+}
+
+func setupLogger(ctx context.Context) context.Context {
+	log := logger.NewLogger()
+	return logger.AddLogger(ctx, log)
+}
--- a/docs/ideas/contextual-logging-example/syslLogger/contextLogger.go
+++ b/docs/ideas/contextual-logging-example/syslLogger/contextLogger.go
@@ -1,0 +1,36 @@
+package logger
+
+import (
+	"context"
+	"errors"
+)
+
+func AddLogger(ctx context.Context, logger Logger) context.Context {
+	return context.WithValue(ctx, key, logger)
+}
+
+func Debug(ctx context.Context, args... interface{}) {
+	log := ctx.Value(key).(Logger)
+	if log == nil {
+		panic(errors.New("Logger does not exist in context"))
+	}
+	log.Debug(args...)
+}
+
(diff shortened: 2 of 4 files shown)


nofun97:
Currently looking at [opentracing](http://opentracing.io/) to see how the logger could be used along with it.

docs/ideas/contextual-logging.md
```diff
@@ -0,0 +1,77 @@
+# Contextual Logging Design Proposal
+
+## Problem
+
+The current logger library, Logrus, makes the use of context quite difficult. In Logrus, the use of context makes use of an intermediate data structure that processes the context data before logging. To achieve this, it is necessary to set up how context is processed through the use of Logrus interface method `Format`. This library aims to make using context with logging easier by standardising log formats and enforces the use of context as part of the effort to use context more in sysl (see issue [#397](https://github.com/anz-bank/sysl/issues/397))
```
anz-gordonj7: can you have another go at this paragraph? English isnt very readable. 

"In Logrus, the use of context makes use of an intermediate data structure that processes the context data before logging." -> I have no idea what this means.

nofun97: added more go, let me know if it's still not clear

docs/ideas/contextual-logging-example/syslLogger/contextLogger.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"context"
+	"errors"
+)
+
+func AddLogger(ctx context.Context, logger Logger) context.Context {
+	return context.WithValue(ctx, key, logger)
+}
+
+func Debug(ctx context.Context, args... interface{}) {
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
func Debug(ctx context.Context, args ...interface{}) {
```

docs/ideas/contextual-logging-example/syslLogger/logger.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type loggerKey struct{}
+
+var key = loggerKey{}
+
+// Partial logger interface
+type Logger interface {
+	Debug(args... interface{})
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	Debug(args ...interface{})
```

docs/ideas/contextual-logging-example/main.go
```diff
@@ -0,0 +1,40 @@
+package main
+
+import (
+	"context"
+	"github.com/anz-bank/sysl/docs/ideas/contextual-logging-example/syslLogger"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

	logger "github.com/anz-bank/sysl/docs/ideas/contextual-logging-example/syslLogger"
```

docs/ideas/contextual-logging.md
nofun97: fixed

nofun97:
Print seems to be your usual Print but it only appears when the verbose flag is given. Though I agree, it might be better to trim it down to Debug, Info, Warn, Error, and Panic. @anzdaddy

nofun97:
After looking through sysl again, it seems that those API are still used. For now, I will build based on this documentation. I will keep the PR open should there be more API changes or any other comments.

nofun97:
To support a log-specific fields, several ways have been designed. (a log-specific field is a log operation that does not modify the logger's fields in the context)

Currently, the solution is to use the API
```go
func Fields(ctx context.Context, fields map[string]interface{}) loggers.Logger {
	return getLogger(ctx).Copy().PutFields(fields)
}
```

Usage would be like `Fields(ctx, map[string]interface{}{"random": "string"}).Debug("Hello there")`. However, this exposes the logger.

Another proposed solution by @anzdaddy is to put the fields in the log API
e.g.
```go
Debug(
  ctx,
  log.Fields("field1", 1),
  log.Fields("field2", 2),
  "this is the message",
  "and more message"
)
```

This does not expose the logger to user but due to the function signature of the logging API `func(ctx context.Context, args ...interface{})`, the function argument won't be the most descriptive and might not be the most ideal solution.

With current API there is a workaround to do a log-specific field without exposing the logger which is
```go
log.Debug(WithField(ctx, "random", "field"), "this is the message")
```
Should this be used, we can remove the `Field` and `Fields` API.

Which would you prefer to use?

anzdaddy:
Slight variation to the example:
```go
var fixedFields = log.Fields(…)
⋮
Debug(ctx,
    fixedFields,
    log.Field("foo", 42),
    "this is the message",
    "and more message"
)
```

anzdaddy:
> LGTM, api is a good base.
> 
> My only question to @anzdaddy is if we really need to many log levels?
> 
> What is really the difference between TRACE and DEBUG? and FATAL/PANIC? Does the logger really need Exit()? What is Print()?
> 
> I would like to trim it down to simply Debug, Info, Warn, Error, and Panic (and even then its probably more than necessary).

I also don't like so many log levels. The catch here is that this isn't just for use within Sysl itself, but also for generated code to use. There shouldn't be a hard dependency. Codegen should be free to depend on whatever it wants, but this is what we plan to expose out of the box.

Perhaps for now we should pare it down to what we consider essential, then discuss whether/how to expand it in Jan.

juliaogris:
https://dave.cheney.net/2015/11/05/lets-talk-about-logging

anzdaddy:
Also, http://peter.bourgon.org/go-in-production/#logging-and-telemetry.

nofun97:
Based on the previous discussion, some changes are proposed.

1. `Print` is to be removed in favour of `Trace`
2. For `Fields` API, the proposed signature is to have change all the logging functionalities signature. For example, `Debug` becomes `Debug(ctx context.Context, message string, fields ...Fields)` or something similar. The fields in that argument are log-specific fields. The idea is to put log message first, restricting it to one message, and putting `Fields` after. This removes the need of the `Field` and `Fields` API which expose the `Logger`.

nofun97:
After further discussion with @Joshcarp it seems @anzdaddy solution might be better because the format variant of the logging API such as `Debugf, Panicf, ...` may have confusing function signature with the proposed `Fields` argument. I would add that, in using @anzdaddy solution, instead of putting `Fields` at the start of `args`, I will put it at the end, so that for the format variant, the format string and the format values can be next to each other.

so instead of
```go
Debugf(
  "%s %s",
  log.Field("field1", 1),
  log.Field("field2", 2),
  "this is a message",
  "this is another message"
)
```

it will be like this

```go
Debugf(
  "%s %s",
  "this is a message",
  "this is another message",
  log.Field("field1", 1),
  log.Field("field2", 2)
)
```

nofun97:
closed due to it being moved to another library https://github.com/anz-bank/pkg

PR #431, opened 2019-11-20
Feature/initial repl support

anz-gordonj7:
Introduces a `repl` command which can execute some simple sysl expressions. The next step is to hook this up to the eval engine so it can work as a live debugger!

This PR is split into 3 commits which will hopefully make reviewing a bit easier

@anz-bank/sysl-developers


sysl2/sysl/cmd_repl.go
```diff
@@ -0,0 +1,28 @@
...
+
+func (p *replCmd) Configure(app *kingpin.Application) *kingpin.CmdClause {
+	cmd := app.Command(p.Name(), "Enter a sysl REPL")
+	return cmd
+}
+
+func (p *replCmd) Execute(args ExecuteArgs) error {
+	s := &eval.Scope{}
+	repl := eval.NewREPL(os.Stdin, os.Stdout)
+	for {
```
springwiz: Its an infinite loop that breaks on error, what if there is no error?

sysl2/sysl/cmd_repl.go
```diff
@@ -0,0 +1,28 @@
...
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+type replCmd struct{}
+
+func (p *replCmd) Name() string            { return "repl" }
+func (p *replCmd) RequireSyslModule() bool { return false }
+
+func (p *replCmd) Configure(app *kingpin.Application) *kingpin.CmdClause {
+	cmd := app.Command(p.Name(), "Enter a sysl REPL")
```
springwiz: Help string "Enter a sysl REPL" can be a little more descriptive.

sysl2/sysl/cmd_repl.go
anz-gordonj7: expected use is user presses q to exit

PR #432, opened 2019-11-20
Fix #430 - check for --help flag before checking empty flags

### LINKED ISSUE #430: Help does not work
Please **do not post any internal, closed source snippets** on this public issue tracker!

### Expected behaviour

`$ go run ./sysl2/sysl import --help`

Should give help on how to use the import command

### Actual behaviour

    ERRO[0000] 'output' value passed is empty
    'format' value passed is empty 
    exit status 1

just try to run the help on any command

anz-gordonj7:
Fixes #430 

If the --help flag is provided on a specific command, but that command has any empty-by-default flags the woould error out instead of showing the help text, fix that.

@anz-bank/sysl-developers

### DIFF
--- a/sysl2/sysl/cmd_runner.go
+++ b/sysl2/sysl/cmd_runner.go
@@ -84,6 +84,11 @@ func EnsureFlagsNonEmpty(cmd *kingpin.CmdClause, excludes ...string) {
 	}
 	fn := func(c *kingpin.ParseContext) error {
 		var errorMsg strings.Builder
+		for _, elem := range c.Elements {
+			if f, _ := elem.Clause.(*kingpin.FlagClause); f != nil && f.Model().Name == "help" {
+				return nil // help requested, don't need to check for empty flags
+			}
+		}
 		for _, f := range cmd.Model().Flags {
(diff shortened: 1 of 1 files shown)


sysl2/sysl/cmd_runner.go
```diff
@@ -84,6 +84,13 @@ func EnsureFlagsNonEmpty(cmd *kingpin.CmdClause, excludes ...string) {
 	}
 	fn := func(c *kingpin.ParseContext) error {
 		var errorMsg strings.Builder
+		for _, elem := range c.Elements {
+			if f, _ := elem.Clause.(*kingpin.FlagClause); f != nil {
+				if f.Model().Name == "help" {
```
nofun97: nitpicky but what about
```go
if f, _ := elem.Clause(*kingpin.FlagClause); f != nil && f.Model.Name == "help"{
    return nil
}
```

nofun97: or is that too long?

anz-gordonj7: indeed, very nitpicky, but maybe slightly moore readable.

PR #440, opened 2019-11-21
Feature/import grammars

anz-gordonj7:
1) complete the grammar->sysl importer
2) fix a couple of importer bugs with regards to sized types
3) Change thee codegen validator to automatically generate a .sysl from the supplied grammar file instead of looking for one (which is likely incorrect or oout of sync anyway)

*.gen.sysl should be removed from docs/examples/everywehre

@anz-bank/sysl-developers


sysl2/sysl/importer/tests-xsd/namespace-simple.sysl
```diff
@@ -12,7 +12,7 @@ testapp [package="package_foo"]:
     # definitions
 
     !type PurchaseOrderType [~xml_root]:
-        ShipTo <: sequence of USAddress(0..2):
+        ShipTo(0..2) <: sequence of USAddress:
```
anzdaddy: What's driving this change? Constraints apply to types, not to fields.

sysl2/sysl/importer/writer.go
```diff
@@ -252,6 +254,9 @@ func (w *writer) writeDefinitions(types TypeList) {
 		switch {
 		case isBuiltInType(t):
 			// do nothing
+		case isUnionType(t):
+			w.writeLines(BlankLine)
+			w.writeUnion(t)
```
anzdaddy: This switch has become a very confused mix of type checks which looks like it could be done as a simple type-assertion switch. An added advantage would be that the switch variable would be of type `*Union`.

sysl2/sysl/importer/writer.go
```diff
@@ -311,6 +322,16 @@ func (w *writer) writeExternalAlias(item Type) {
 		PushIndent, aliasType, PopIndent)
 }
 
+func (w *writer) writeUnion(item Type) {
```
anzdaddy: Parameter type should be `*Union`.

sysl2/sysl/importer/tests-xsd/namespace-simple.sysl
anz-gordonj7: The original wasn't valid according to the parser

sysl2/sysl/importer/writer.go
anz-gordonj7: Disagree, those is*Type() funcs are slightly coomplicated and the ordering matters, would prefer to leave as is

PR #442, opened 2019-11-25
Fix function/view calls, add 'l' to list views

anz-gordonj7:
Hook up the debugger to the codegen command (use --debugger)

attempts to add more source context (adds the end and in some cases the expression text),
use 'l' to list the views in the current app, breakpoints still to come

@anz-bank/sysl-developers


sysl2/sysl/eval/debugger.go
```diff
@@ -149,40 +173,48 @@ func parseExpression(text string, scope *Scope) string {
 
 	p := parser.NewSyslParser(stream)
 	p.AddErrorListener(&errorListener)
-	p.AddParseListener(listener)
+	p.BuildParseTrees = true
+
+	defer func() {
+		recover()
```
golangcibot: Error return value of `recover` is not checked (from `errcheck`)

sysl2/sysl/parse/utils.go
```diff
@@ -15,13 +15,31 @@ type sourceCtxHelper struct {
 	filename string
 }
 
-func (s sourceCtxHelper) Get(ctx *antlr.BaseParserRuleContext) *sysl.SourceContext {
+func (s sourceCtxHelper) Get(ctx *antlr.BaseParserRuleContext, withText ...bool) *sysl.SourceContext {
```
gkanz: The bool to include extra data makes me feel uneasy because:
1. Reading the source code it's not clear what the trailing `true` is doing.
2. If additional things that can be returned in various combinations are added the code gets messy. 

Perhaps a GetWithText wrapper method?

anz-gordonj7: fair enough, will change

anz-gordonj7:
#464 supersedes this

PR #445, opened 2019-11-26
Allow a second scope variable to be used with transforms:

anz-gordonj7:
Maps will use "key, value" if 2 scope vars are used
Lists/Sets will use "index, value" if 2 scope vars are used

No change if 0 or 1 scopevar are used.
```
!view foo(app <: sysl.App) -> string:
   app.endpoints -> (name, ep:
      # name is the value that would have been .key
      # ep is the value that would have been .value
   )

!view bar(someList <: sequence of string) -> string:
   someList -> (i, item:
     # i is the list index (0 based)
     # item is the list item
   )
```
Fixes # .

-
-

@anz-bank/sysl-developers


gkanz:
I like that it is like the Go way of doing things!

Does this break existing code? What will happen for example if two values are returned but there is only one there to catch it?

I think accessing "i" in the second example may lead to some fragile sysl code as order will change with the contents of the input sysl file. Would the sequence index,value returns also be implemented for sets?

anz-gordonj7:
> Does this break existing code? What will happen for example if two values are returned but there is only one there to catch it?
 
The transform execution checks how many scope params are expected (same as go), so any existing code will be unchanged (it will always default to passing the value for lists/sets or the implicit map for map types)


> I think accessing "i" in the second example may lead to some fragile sysl code as order will change with the contents of the input sysl file. Would the sequence index,value returns also be implemented for sets?

sequences/sets are both implemented as go slices, so whatever order they are in the slice is the order the view will receive them in. I could make it so sets are ordered (like maps), but in either case the order should remain stable (it comes down to the parse ordering, not map inseertion)

springwiz:
This change needs the sysl grmmar to fixed up as well. So that:
let x = list -> <sequence of ListParams> (indx, p: OR
let x = map -> <sequence of MapParams> (key,p:
passes the sysl compile.

PR #458, opened 2019-11-28
Fix the grammar importer

anz-gordonj7:
Fixes the grammar importer, generates simpler type names

@anz-bank/sysl-developers


sysl2/sysl/importer/grammar.go
```diff
@@ -31,137 +35,236 @@ func LoadGrammar(args OutputData, grammarInput string, logger *logrus.Logger) (o
...
-				Target: &SyslBuiltIn{name: StringTypeName},
+func (rp *ruleProcessor) buildSequence(s *ebnfGrammar.Sequence) string {
+	var fields []term
+	for _, t := range s.Term {
+		fields = append(fields, rp.buildTerm(t))
+	}
+	if len(fields) == 1 && fields[0].field == nil {
+		rp.deps.Insert(fields[0].name)
+		return fields[0].name
+	} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

anz-gordonj7:
@gkanz https://gist.github.com/anz-gordonj7/3bc3aa93c962dd66300f84dc52026077 has the manually created go.gen.sysl and the version created by the importer.

Key differences:

* Expression rule, in generated is a simple union (line 45), is manual (line 223) it is a type with a single member which is the choice union. This is breaking the validator.

* generated version contains the array size spec which more closely matches the input grammar (unsure if this breaks the validator)

* I personally think the generated type names are nicer than the __Choice_* names in the manual version. example: `!type __Choice_Combination_MethodSpec_MethodName_Signature:` (line 132) and with `__MethodSpec_01` (line 150 of generated)

PR #464, opened 2019-12-05
Fix debug evaluation

anz-gordonj7:
Fixes debugger evaluation and adds function calls, and a --debugger option to the command line.

This is really just #442 rebased onto master which was difficult on that branch

@anz-bank/sysl-developers


pkg/eval/debugger_test.go
```diff
@@ -156,7 +160,23 @@ func TestREPL_DumpScope(t *testing.T) {
...
-	require.Contains(t, result, "a:  s:\"hello\"")
-	require.Contains(t, result, "b:  b:false")
-	require.Contains(t, result, "c:  null:<>")
+	for k, v := range *scope {
+		require.Contains(t, result, fmt.Sprintf("%s:  %s", k, unaryString(v).GetS()))
+	}
+}
+
+func TestREPL_CallView(t *testing.T) {
+	filename := "../../cmd/sysl/tests/transform1.sysl"
```
Joshcarp: I like the idea of having a const called syslroot or something so there aren't so many dots everywhere

Issue #478, opened 2019-12-16
Simplify package main


Joshcarp:
`package main` currently has 39 files in it (including test files)
This should be simplified so that all that `package main` has no feature work in it

PR #479, opened 2019-12-16
[WIP] add command mod to support sysl modules

### LINKED ISSUE #415: Support Sysl imports across repos
Sysl definitions living in multiple independent repos will need to refer to each other.

Implement a mechanism to refer to and access external repo references. Perhaps take inspiration from go modules. Perhaps delegate the problem to bazel.

ChloePlanet:
Fixes #415 

- sysl mod init
- sysl mod tidy
- sysl mod vendor
- sysl mod clean

@anz-bank/sysl-developers

### DIFF
--- a/cmd/sysl/cmd_mod.go
+++ b/cmd/sysl/cmd_mod.go
@@ -1,0 +1,19 @@
+package main
+
+import (
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+type modCmd struct {}
+
+func (p *modCmd) Name() string            { return "mod" }
+func (p *modCmd) RequireSyslModule() bool { return false }
+
+func (p *modCmd) Configure(app *kingpin.Application) *kingpin.CmdClause {
+	cmd := app.Command(p.Name(), "Provide access to operations on modules")
+	return cmd
+}
+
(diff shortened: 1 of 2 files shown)


cmd/sysl/cmd_mod.go
```diff
@@ -0,0 +1,19 @@
+package main
+
+import (
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+type modCmd struct {}
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
type modCmd struct{}
```

cmd/sysl/cmd_mod.go
```diff
@@ -0,0 +1,19 @@
+package main
```
springwiz: Lets not add this to main package and figure out a separate package for this.

ChloePlanet: I think @Joshcarp is working on this(#478 ), so I just started as it is. I will sync up the code with his changes after his part's done.

anz-gordonj7: No.... main will continue to have these cmd files, they should just call into pkg/<something> to actually do the work. think of main as the frontend and pkg/ as the backend

PR #483, opened 2019-12-17
Initial work to import openapi 3

anz-gordonj7:
Implements OpenAPI 3.0 as a new importer.

Unfortunately I havnt found a go pkg which nartivly handles both swagger and openapi so duplicating lots of work :(

@anz-bank/sysl-developers


pkg/importer/openapi.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+}
+
+func (l *loader) initEndpoints() []MethodEndpoints {
+	epMap := map[string][]Endpoint{}
+
+	l.initGlobalParams()
+
+	for path, item := range l.spec.Paths {
+		ops := map[string]*openapi3.Operation{
```
springwiz: Are these the only operations supported in Open API, there is no Http OPTIONS and HEAD?

pkg/importer/importer_test.go
```diff
@@ -59,6 +59,15 @@ func TestLoadSwaggerFromTestFiles(t *testing.T) {
 	})
 }
 
+func TestLoadOpenApiFromTestFiles(t *testing.T) {
```
springwiz: Possibly target higher code coverage by splitting up the tests.

Joshcarp:
Only one commit for all of this?

pkg/importer/importer_test.go
anz-gordonj7: 93.8% coverage on just openapi.go from the added tests... whats your problem exactly?

pkg/importer/openapi.go
anz-gordonj7: If thesee end up being required they can be added later

pkg/importer/importer_test.go
springwiz: I thought of splitting them considering only one test entity at a time. Thus you have test cases for importing open-api Endpoints, Types so on and so forth. But the coverage is good as is, so it's fine.

pkg/importer/openapi.go
springwiz: ok

PR #490, opened 2019-12-18
Fix installation docs

juliaogris:
Fix the installation command
Reword various while at it. Hopefully this is better - not sure though

PR: https://deploy-preview-490--adoring-liskov-4cef13.netlify.com/docs/installation/   
Before: https://sysl.io/docs/installation/

@anz-bank/sysl-developers

### DIFF
--- a/docs/website/content/docs/installation.md
+++ b/docs/website/content/docs/installation.md
@@ -10,6 +10,8 @@ toc: true
 
 Sysl is a CLI (Command Line Interface) that executes with the `sysl` command.
 
+Install it with
+
     GO111MODULE=on go get github.com/anz-bank/sysl/cmd/sysl@v0.4.0
 
 Note: Do NOT run it from inside a Go source directory that is module enabled,
@@ -17,11 +19,11 @@ otherwise it gets added to go.mod/go.sum.
 
 ## Prerequisites
 
-- [Go](https://golang.org)
-- [PlantUML](https://hub.docker.com/r/plantuml/plantuml-server/) server for diagram generation for use if using the [external service](http://www.plantuml.com/plantuml/) is not appropriate
+- [Go 1.13](https://golang.org/doc/install)
+- PlantUML environment variable for diagram generation:
+  `export SYSL_PLANTUML=http://www.plantuml.com/plantuml`
 
-## Setting PlantUML Environment variable
-
-In order to be able to generate diagrams the `SYSL_PLANTUML` Environment variable needs to be set
-
-    export SYSL_PLANTUML=http://www.plantuml.com/plantuml
+If the external PlantUML service is not suitable for your use case, you can run
+a local instance of the [PlantUML
+server](https://hub.docker.com/r/plantuml/plantuml-server/) using the docker
+image and point the `SYSL_PLANTUML` environment variable to that instance.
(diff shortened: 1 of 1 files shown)


docs/website/content/docs/installation.md
```diff
@@ -7,17 +7,20 @@ draft: false
 bref: "Sysl can be installed on Windows, MacOS and Linux - follow this guide"
 toc: true
 ---
-Sysl is a CLI (Command Line Interface) that excecutes with the `sysl` command. 
 
-`go get -v github.com/anz-bank/sysl/cmd/sysl`
+Sysl is a CLI (Command Line Interface) that executes with the `sysl` command.
```
juliaogris: This still needs to be reworded imo - i'll leave to the native speakers if they agree.

docs/website/content/docs/installation.md
```diff
@@ -10,7 +10,7 @@ toc: true
 
 Sysl is a CLI (Command Line Interface) that excecutes with the `sysl` command.
 
-    go get -v github.com/anz-bank/sysl/cmd/sysl
+    GO111MODULE=on go get github.com/anz-bank/sysl/cmd@v0.4.0
```
camh-anz: should be `.../sysl/cmd/sysl@v0.4.0`. i.e. needs more sysl

camh-anz: This also needs a note to NOT run it from inside a Go source directory that is module enabled, otherwise it gets added to `go.mod`/`go.sum`. I just tested with a fresh module: `mkdir xxx; cd xxx; go mod init xxx; GOMODULE111=on go get ...` and it updated the go modules. From a quick look of the help, I can't see any way to tell `go get` not to do this. sigh.

docs/website/content/docs/installation.md
camh-anz: It needs to be re-worded, but I'm not quite sure what it needs to say. Is Sysl more than just the cli? is some library you link against? or is the `sysl` CLI the only entrypoint into SYSL?

docs/website/content/docs/installation.md
```diff
@@ -10,15 +10,17 @@ toc: true
...
-
-## Setting PlantUML Environment variable
-
-In order to be able to generate diagrams the `SYSL_PLANTUML` Environment variable needs to be set
+- [Go 1.13](https://golang.org/doc/install)
+- PlantUML environment variable for diagram generation:
+  `export SYSL_PLANTUML=http://www.plantuml.com/plantuml`
 
-    export SYSL_PLANTUML=http://www.plantuml.com/plantuml
+If the unauthenticated, external [PlantUML
```
camh-anz: I don't think "unauthenticated" is the right word here. I would just say something like:

If the external PlantUML service is not suitable for your use case, you can run a local instance of the PlantUML server using the docker image and point the `SYSL_PLANTUML` environment variable to that instance.

docs/website/content/docs/installation.md
juliaogris: Done 🤦‍♀

docs/website/content/docs/installation.md
juliaogris: If you want you can say it is also a programming language - right?
And as such there have been various IDE/automation attempts around it -  e.g. build it into a website or so.
It is also intended to be "easily extensible" I think - like `git-pluginname` or so, not sure if any of that has happened yet.
I'll just leave it for now.

docs/website/content/docs/installation.md
juliaogris: Done. Thanks for your clarity.

docs/website/content/docs/installation.md
juliaogris: Done.

PR #495, opened 2020-01-01
Feature/refactor listener

anz-gordonj7:
Some structural refactoring of the antlr parser listener to allow for parser simplification to follow...

BasicallyI noticed that the code was reusing struct fields in multiple places for different purposes which meant makes working in this codebase hell.

* Splits usages of `s.typename []string` into distinct fields for each of its uses (including:)
      1. keeping track of names used to build up references to types
      1. endpoint paths
      1. collector paths
      1. view names(!)
* Adds a simple stack type to replace the multiple ad-hock stacks that are used with the `typename` and `url_prefix` fields
* Replace the repeated map lookup to get the current modules app with a helper func

@anz-bank/sysl-developers

Reviewers: I would look at the commits indivually instead of the full diff, each is self contained.

### DIFF
--- a/pkg/parse/utils.go
+++ b/pkg/parse/utils.go
@@ -100,7 +100,7 @@ func primitiveFromNativeDataType(native antlr.TerminalNode) (*sysl.Type_Primitiv
 	return &sysl.Type_Primitive_{Primitive: primitiveType}, constraint
 }
 
-type pathStack struct {
+type PathStack struct {
 	sep   string
 	parts []string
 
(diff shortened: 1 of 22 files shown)


pkg/parse/utils.go
```diff
@@ -99,3 +99,51 @@ func primitiveFromNativeDataType(native antlr.TerminalNode) (*sysl.Type_Primitiv
...
+func (p *PathStack) Reset() string {
+	p.parts = []string{}
+	return p.update()
+}
+
+// Parts() clones the path items into a new slice so that it is safe to store
+func (p PathStack) Parts() []string {
+	out := make([]string, len(p.parts))
+	copy(out, p.parts)
+	return out
```
anzdaddy:
```go
return append([]string{}, p.parts...)
```

pkg/parse/utils.go
```diff
@@ -99,3 +99,51 @@ func primitiveFromNativeDataType(native antlr.TerminalNode) (*sysl.Type_Primitiv
 	}
 	return &sysl.Type_Primitive_{Primitive: primitiveType}, constraint
 }
+
+type PathStack struct {
+	sep   string
+	parts []string
```
anzdaddy: nit: I suspect the code will end up simpler without `parts`. Since most calls invoke `p.update()`, it might also be more efficient.

anz-gordonj7: `p.update()` was really just premature optimisation and should/could probably be removed.

I'm actually hoping to remove the need for this completely in follow up PR's.

pkg/parse/utils.go
anz-gordonj7: done

PR #497, opened 2020-01-06
unify type/references creation

anz-gordonj7:
This PR pulls out almost all the type parsing into a seperate child parser as a further step to simplifying listener_impl.go. The reason for this is:
1) Ensure consistency around usage of types in the parser, anywhere which expects a type should allow any type of type.
2) Centralise all type creation so that TypeRef types can be removed from the parsed sysl file (so eval code can use the real sysl.Type instead of guessing the type via strings.

@anz-bank/sysl-developers


pkg/parse/tests/foreign_import_swagger.sysl.golden.textpb
```diff
@@ -131,6 +131,16 @@ apps: <
           query_param: <
             name: "depth"
             type: <
+              type_ref: <
```
sahejsingh: These changes ( golden.textpb changes without corresponding .sysl changes) are regressions and not compatible with legacy python.
Need to be looked at more closely.

anz-gordonj7: compatibility is not a priority, anz-bank/sysl-legacy is where that belongs.

pkg/parse/resolvetyperef_test.go
```diff
@@ -0,0 +1,173 @@
...
+		Ref: &sysl.Scope{
+			Appname: &sysl.AppName{
+				Part: t.appname,
+			},
+			Path: t.path,
+		},
+	}
+}
+
+func (t typereftestdata) Run(tt *testing.T, mod *sysl.Module) {
```
joshcarp: Comment

pkg/parse/typeparser.go
```diff
@@ -0,0 +1,127 @@
...
+		Type: &sysl.Type_TypeRef{
+			TypeRef: &sysl.ScopedRef{
+				Context: syslContext,
+				Ref:     ref,
+			},
+		},
+	}
+}
+
+func (t *typeParser) EnterName_str(ctx *parser.Name_strContext) {
```
joshcarp: Comment

pkg/parse/typeparser.go
```diff
@@ -0,0 +1,127 @@
...
+	*t.namestrTarget = append(*t.namestrTarget, ctx.GetText())
+}
+
+func (t *typeParser) EnterUser_defined_type(ctx *parser.User_defined_typeContext) {
```
joshcarp: Comment

pkg/parse/typeparser.go
```diff
@@ -0,0 +1,127 @@
...
+	t.namestrTarget = &t.typePath
+}
+
+func (t *typeParser) EnterReference(ctx *parser.ReferenceContext) {
```
joshcarp: Comment

pkg/parse/typeparser.go
```diff
@@ -0,0 +1,127 @@
...
+	t.namestrTarget = &t.appName
+}
+
+func (t *typeParser) ExitApp_name(ctx *parser.App_nameContext) {
```
joshcarp: Comment

anzdaddy:
@joshcarp Let me have a go at changing your mind. In Go, comments are predominantly used for documenting exported symbols. Everywhere else, they should be used as sparingly as possible, mostly relying on the clarity of the code itself to document the intent.

In this instance, the symbols are exported only by necessity because they implement the methods of the `SyslParserListener` callback interface and is not intended for consumption by a developer.

anz-gordonj7:
@joshcarp are you asking for comments on those functions because "they are exported" or because you the code doesnt make it clear what it is doing?

PR #500, opened 2020-01-07
Import sysl files across repos

### LINKED ISSUE #415: Support Sysl imports across repos

ChloePlanet:
Fixes #415  .

- Try to download imported files from internet based on `go modules` when it cannot be found locally in parsing process
- Corresponding test

@anz-bank/sysl-developers


ChloePlanet:
Considering replace `go modules` with [OCI](https://www.opencontainers.org/blog/2019/09/10/new-oci-artifacts-project)

pkg/mod/load.go
```diff
@@ -0,0 +1,93 @@
...
+		dir = filepath.Dir(dir)
+	}
+
+	return errors.New("No such module")
+}
+
+func GetExternalFile(filename string) (string, error) {
+	err := goGetByFilepath(filename)
+	if err != nil {
+		return filename, fmt.Errorf("%s not found\n", filename)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

pkg/mod/load.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	if err = LoadAll(); err != nil {
+		return filename, fmt.Errorf("error loading modules: %s\n", err.Error())
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

pkg/mod/load.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	mod := GoMods.GetByFilepath(filename)
+	if mod == nil {
+		return filename, fmt.Errorf("error finding module of file %s\n", filename)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

pkg/mod/util.go
```diff
@@ -0,0 +1,49 @@
...
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+func runGo(
+	ctx context.Context,
+	stdout io.Writer,
+	args ...string) error {
```
golangcibot: unnecessary leading newline (from `whitespace`)

pkg/mod/util.go
```diff
@@ -0,0 +1,49 @@
...
+		if !ok {
+			return errors.Errorf("failed to execute 'go %v': %s %T", args, err, err)
+		}
+
+		// Too old Go version
+		if strings.Contains(stderr.String(), "flag provided but not defined") {
+			return nil
+		}
+		return errors.Errorf("go command failed: %s", stderr)
+
```
golangcibot: unnecessary trailing newline (from `whitespace`)

ChloePlanet:
To test the example:
`./sysl validate ./demo/examples/Modules/model_with_deps.sysl`

The external module must be a go module now which means it should have `go.mod` file under the working directory.

pkg/mod/util.go
```diff
@@ -0,0 +1,44 @@
...
+	"context"
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+func runGo(ctx context.Context, stdout io.Writer, args ...string) error {
```
anz-gordonj7: be careful with variable names... stdout and stderr have specific meaning. this var name should be `out`, and stderr below shuold probably be something like `errbuf`.

Also, I'm not sure this should be writing to os.Stderr at all, I'd hook cmd.Stderr up too stdout

anz-gordonj7: This shuold log at debug or trace level the full command being executed

pkg/parse/fs.go
```diff
@@ -13,6 +13,8 @@ func fileExists(filename string, fs afero.Fs) bool {
 	if err != nil {
 		return false
 	}
+	defer f.Close()
```
anz-gordonj7: good catch!

pkg/parse/parse.go
```diff
@@ -2,11 +2,13 @@ package parse
```
anz-gordonj7: Revert this whole file!

Finding files isnt the responsibility of the parser, it is the responsibility of the `afero.Fs`, I would move all this remote loading implementation details to a new afero.Fs implementation which is backed by some other fs, which tried to download if the file doesnt exist. something like:

```
type FsWithModSupport struct {
    fs afero.Fs // the real filesystem to use for file accesses
    ... // whatever is actually needed to run pkg/mod
}

func (fs *FsWithModSupport) OpenFile(filename string) { // whatever the API actually is
   1) attempt to load the file from fs.fs
   2) if it doesnt exist and it looks like a url do the mod download
   3) shuold now be able to load the file from fs.fs
}
```

An advantage of this is testing the fs can be done much more easily, and it can be swapped out if the user wants to forbid downloads (can be a security vulnerability)

pkg/mod/mod.go
```diff
@@ -0,0 +1,40 @@
+package mod
+
+import (
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+type goModule struct {
```
anz-gordonj7: that fact we are using `go mod` is an implementation detail, users of pkg/mod shouldnt care what type of mod it is, as long as it provides .sysl files. this could just be `module` or similar

pkg/mod/mod.go
```diff
@@ -0,0 +1,40 @@
...
+	Path     string     // module path
+	Version  string     // module version
+	Versions []string   // available module versions (with -versions)
+	Replace  *goModule  // replaced by this module
+	Time     *time.Time // time version was created
+	Update   *goModule  // available update, if any (with -u)
+	Main     bool       // is this the main module?
+	Indirect bool       // is this module only an indirect dependency of main module?
+	Dir      string     // directory holding files for this module, if any
+	GoMod    string     // path to go.mod file for this module, if any
```
anz-gordonj7: These comments are not really available when someone is reading code later on that uses this struct, can you think of better field names so these comments are not needed.

i.e, lets say we have a variable of this type called `mod`, `mod.Path` is a fairly obvious name, `mod.Versions` and `mod.Version` is confusing (why have 2?).

`mod.Replace`? does this mean the mod is replaced by the one pointed to, or is this one replacing the pointed to one?

why `Path` and `Dir`?

Getting an API right is hard and extremely important, practice (and having to use bad API) will make this easy eventually.


Additionally, almost all of these don't appear to be used, YAGNI (look it up if you havnt seen that acronym before)

pkg/mod/mod.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+type goModules []*goModule
```
anz-gordonj7: I'm not sure I understand the point of this type, or how it is supposed to be used.

Would it be better to make the func below `func GetByFilepath(mods goModules, path string) *goModule` which would mean the type isnt needed anymore? Do you expect users to use goModules as a slice or an object? 
`mods.Add(mod)` is nicer than `mods = append(mods, mod)`

pkg/mod/mod.go
```diff
@@ -0,0 +1,40 @@
...
+
+var GoMods goModules
```
anz-gordonj7: you should be getting a lint error from this, globals are forbidden, they serve almost no purpose and should likewise almost never be used! I don't see a valid reason to allow this one.

pkg/mod/load.go
```diff
@@ -0,0 +1,93 @@
...
+	"io"
+	"io/ioutil"
+	"path/filepath"
+	"regexp"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+func LoadAll() error {
```
anz-gordonj7: a func which takes no arguments and is not a method receiver means global variables, which is never safe. I'm sure we can come up with a version of this without the need for globals.

pkg/mod/mod.go
ChloePlanet: Thanks for the detailed feedback! I'll rethink it.

pkg/mod/fs.go
```diff
@@ -0,0 +1,80 @@
...
+	return &ModSupportedFs{fs: fs}
+}
+
+func (fs *ModSupportedFs) Open(name string) (afero.File, error) {
+	f, err := fs.fs.Open(name)
+	if err == nil {
+		return f, nil
+	}
+
+	mod, err := Find(name)
```
golangcibot: undeclared name: `Find` (from `typecheck`)

pkg/parse/parse.go
```diff
@@ -109,10 +112,25 @@ func (p *Parser) Parse(filename string, fs afero.Fs) (*sysl.Module, error) {
...
 		logrus.Debugf("Parsing: " + filename)
+
 		fsinput, err := newFSFileStream(filename, fs)
 		if err != nil {
-			return nil, Exitf(ImportError, fmt.Sprintf("error parsing %#v: %v\n", filename, err))
+			if _, ok := err.(*os.PathError); !ok {
+				return nil, Exitf(ImportError, fmt.Sprintf("error parsing %#v: %v\n", filename, err))
+			}
+
+			filename, err = mod.GetExternalFile(filename)
```
golangcibot: GetExternalFile not declared by package mod (from `typecheck`)

pkg/mod/mod_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	var testMods goModules
+	testMods = append(testMods, &goModule{Path: "modulepath"})
+	assert.Nil(t, testMods.GetByFilepath("another_modulepath/filename"))
+}
+
+func TestGetByFilepathWithNilMods(t *testing.T) {
+	t.Parallel()
+
+	var testMods goModules
```
golangcibot: undeclared name: `goModules` (from `typecheck`)

pkg/mod/load_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+	mod := GoMods.GetByFilepath(filename)
+	assert.Equal(t, filepath.Join(mod.Dir, "/demo/examples/Modules/deps.sysl"), importFilename)
+}
+
+func TestGetExternalFileWithWrongPath(t *testing.T) {
+	t.Parallel()
+
+	wrongpath := "wrong_file_path/deps.sysl"
+	importFilename, err := GetExternalFile(wrongpath)
```
golangcibot: undeclared name: `GetExternalFile` (from `typecheck`)

pkg/mod/load_test.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+func TestGetExternalFile(t *testing.T) {
+	t.Parallel()
+
+	filename := "github.com/anz-bank/sysl/demo/examples/Modules/deps.sysl"
+	importFilename, err := GetExternalFile(filename)
+	assert.Nil(t, err)
+
+	mod := GoMods.GetByFilepath(filename)
```
golangcibot: undeclared name: `GoMods` (from `typecheck`)

pkg/mod/fs.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/anz-bank/sysl/pkg/syslutil"
+	"github.com/spf13/afero"
+)
+
+type ModSupportedFs struct {
```
golangcibot: type name will be used as mod.ModSupportedFs by other packages, and that stutters; consider calling this SupportedFs (from `golint`)

pkg/parse/parse.go
```diff
@@ -99,20 +100,27 @@ func (p *Parser) Parse(filename string, fs afero.Fs) (*sysl.Module, error) {
...
 	}
+
 	imported := map[string]struct{}{}
 	listener := NewTreeShapeListener()
 
 	source := importDef{
 		filename: filename,
 	}
 
+	if mod.SyslModules {
```
anz-gordonj7: This could be added all the way up in cmd/sysl/cmd_runner.go in Run()

pkg/mod/fs_test.go
```diff
@@ -0,0 +1,48 @@
...
+	f, err := fs.Open(filename)
+	assert.Nil(t, err)
+	assert.Equal(t, "deps.sysl", filepath.Base(f.Name()))
+}
+
+func TestOpenRemoteFile(t *testing.T) {
+	t.Parallel()
+
+	filename := "github.com/ChloePlanet/sysltestpub/pubbanana.sysl"
+	fs := NewFs(afero.NewOsFs())
```
anz-gordonj7: tests should never touch the host filesystem, if it needs to write it should be writing to a MemFs()

pkg/mod/fs_test.go
```diff
@@ -0,0 +1,48 @@
...
+	f, err := fs.Open(filename)
+	assert.Nil(t, err)
+	assert.Equal(t, "pubbanana.sysl", filepath.Base(f.Name()))
+}
+
+func TestOpenRemoteFileFailed(t *testing.T) {
+	t.Parallel()
+
+	filename := "github.com/wrong/sysltestpub/pubbanana.sysl"
```
anz-gordonj7: ditto above

pkg/mod/fs.go
```diff
@@ -0,0 +1,80 @@
...
+	mod, err := Find(name)
+	if err != nil {
+		return nil, err
+	}
+	relpath, err := filepath.Rel(mod.Name, name)
+	if err != nil {
+		return nil, err
+	}
+
+	return syslutil.NewChrootFs(afero.NewOsFs(), mod.Dir).Open(relpath)
```
anz-gordonj7: This is a bit tricky. I'm not super comfortable with this because it opens up the entire filesystem for access, but yeah, the filesystem which the modules need to come from likely will be different from the chroot the module itself comes from.

I think the solution is to pass 2 afero.Fs into this filesystem, one which is used for the non-module files, the second is for the modules,  this also allows for easy testing (you can easily test that files were written to a MemFs in the right place)

pkg/mod/module_test.go
```diff
@@ -0,0 +1,59 @@
...
+	assert.Nil(t, testMods.GetByFilename("modulepath/filename"))
+}
+
+func TestFind(t *testing.T) {
+	t.Parallel()
+
+	filename := "github.com/anz-bank/sysl/demo/examples/Modules/deps.sysl"
+	mod, err := Find(filename)
+	assert.Nil(t, err)
+	assert.Equal(t, "github.com/anz-bank/sysl", mod.Name)
```
anz-gordonj7: what happened to the rest of the filename?

pkg/mod/gomod.go
```diff
@@ -0,0 +1,112 @@
...
+	Dir  string
+}
+
+func goGetByFilepath(filename string) error {
+	dir := filepath.Dir(filename)
+	re := regexp.MustCompile(`^\.|\.\.|/$`)
+
+	for !re.MatchString(dir) {
+		logrus.Debug(dir)
+		err := goGet(dir)
```
anz-gordonj7: where does `go get` dump the new files?

pkg/mod/fs.go
```diff
@@ -0,0 +1,80 @@
...
+func (fs *Fs) Mkdir(name string, perm os.FileMode) error {
+	return fs.fs.Mkdir(name, perm)
+}
+
+func (fs *Fs) MkdirAll(path string, perm os.FileMode) error {
+	return fs.fs.MkdirAll(path, perm)
+}
+
+func (fs *Fs) OpenFile(name string, flag int, perm os.FileMode) (afero.File, error) {
+	return fs.fs.OpenFile(name, flag, perm)
```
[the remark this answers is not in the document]
ChloePlanet: Sure!

pkg/parse/parse.go
ChloePlanet: Cool. Thanks for the tip

pkg/mod/module.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("%s not found", name)
+	}
+
+	var modules Modules
+	if err = modules.Load(); err != nil {
+		return nil, fmt.Errorf("error loading modules: %s", err.Error())
+	}
+
+	if mod == nil {
```
golangcibot: undeclared name: `mod` (from `typecheck`)

pkg/mod/module.go
```diff
@@ -0,0 +1,55 @@
...
+		return nil, fmt.Error("modules list is empty")
```
golangcibot: Error not declared by package fmt (from `typecheck`)

pkg/mod/module.go
```diff
@@ -0,0 +1,55 @@
...
+type Modules []*Module
+
+func (m *Modules) Add(item *Module) {
+	*m = append(*m, item)
+}
+
+func (m *Modules) GetByFilename(p string) *Module {
+	p = filepath.Clean(p)
+
+	for _, m := range m {
```
golangcibot: cannot range over m (variable of type *Modules) (from `typecheck`)

pkg/mod/gomod.go
```diff
@@ -0,0 +1,112 @@
...
+	for {
+		m := &goModule{}
+		if err := dec.Decode(m); err != nil {
+			if err == io.EOF {
+				break
+			}
+			return errors.Wrap(err, "failed to decode modules list")
+		}
+
+		m.Add(&Module{
```
golangcibot: m.Add undefined (type *goModule has no field or method Add) (from `typecheck`)

pkg/mod/fs.go
ChloePlanet: I agree with it. But the problem here is that the root of the fs is `mod.Dir`, which is mutable for every different module. It cannot be confirmed when `cmdRunner.Run`. Do you have any ideas?

pkg/mod/gomod.go
ChloePlanet: What `go get` does is downloading new files to `$GOPATH/pkg/mod` and rewriting `go.mod` & `go.sum`.

pkg/mod/module_test.go
ChloePlanet: `Find` finds go module of the given file. `mod.Name` is the name of related go module. The downloaded go module doesn't know the accurate file path.

pkg/mod/fs.go
anz-gordonj7: Doesn't `go get` download the files into a common root directory? so `import github.com/somewhere/foo.sysl` would end up in `<known location>/github.com/somewhere`? The fs for that could then be a chrootFS at <known location>

pkg/mod/gomod.go
anz-gordonj7: We should change the GOPATH environment variable of the `go get` call so we control the location. i.e the `.sysl/ ` directory if found, or the `SYSLPATH` envvar

pkg/mod/fs.go
ChloePlanet: `go get` downloads the files into `$GOPATH/pkg/mod/github.com/org/repo@version`. Golang hard codes `$GOPATH + "pkg/mod"`, but we don't know `repo@version` at first.

cmd/sysl/cmd_runner.go
```diff
@@ -25,17 +26,22 @@ type cmdRunner struct {
 func (r *cmdRunner) Run(which string, fs afero.Fs, logger *logrus.Logger) error {
 	if cmd, ok := r.commands[which]; ok {
 		if cmd.Name() == which {
-			var mod *sysl.Module
+			var module *sysl.Module
 			var err error
 			var appName string
 			var mods []*sysl.Module
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)

PR #502, opened 2020-01-08
added sysl-safe names checker and converter

### LINKED ISSUE #499: importing query params with hyphens generate bad sysl
Attempting to import a swagger/openAPI spec where a query parameter contains a '-' causes the generated sysl to contain the - also, which is invalid.

example:
```
- name: cust-nums
     type: string
     required: true
     in: query
```

generated out to `?cust-nums=string&....` The python importer would convert that to `custNums=string` which is now valid sysl but no longer follows the spec.

nofun97:
Fixes #499 .

- Added a functionality to check unsafe parameter names and convert it

currently it checks for hyphens and spaces only.

@anz-bank/sysl-developers


pkg/importer/openapi.go
```diff
@@ -337,10 +337,14 @@ func (l *loader) buildParams(params openapi3.Parameters) Parameters {
 }
 
 func (l *loader) buildParam(p *openapi3.Parameter) Param {
+	name := p.Name
+	if hasToBeGoSafe(p.In) {
```
anz-gordonj7: this actually needs to be sysl safe, not go safe. the - breaks the sysl parser

pkg/importer/openapi.go
```diff
@@ -349,3 +353,32 @@ func (l *loader) buildParam(p *openapi3.Parameter) Param {
 		In: p.In,
 	}
 }
+
+func hasToBeGoSafe(in string) bool {
+	return strings.ToLower(in) == "query"
+}
+
+func convertToGoSafe(name string) string {
```
anz-gordonj7: this func is a bit gross to be honest. You could probably make it much simpler by doing `strings.Split(name, "-")` and then rejoining with some smarts.

pkg/importer/tests-swagger/EXAMPLE_SWAGGER_UNSAFE_PARAM_NAME.yaml
```diff
@@ -0,0 +1,105 @@
+swagger: "2.0"
```
anz-gordonj7: EXCELLENT! very thorough testing

pkg/importer/openapi.go
nofun97: i thought of that, but i think with the current function we can support more unsafe characters later if needed, what do you think?

pkg/importer/openapi.go
nofun97: changed to sysl safe

pkg/importer/openapi.go
anz-gordonj7: its fine

PR #505, opened 2020-01-10
Fix issue 475.

ericzhang6222:
Fixes # 475 .

- Issue 475 is caused by statement 'PackageClause: 'package' PackageName '\n';', so remove unnecessary semicolon.  

@anz-bank/sysl-developers


ericzhang6222:
> this also neeeds to be on the internal server-lib/ repo

It is correct in server-lib/repo as https://github.service.anz/sysl/server-lib/blob/master/codegen/grammars/go.gen.g, it is 'PackageClause: 'package' PackageName '\n';'.

PR #516, opened 2020-01-16
fixes #509, swagger now works with remote refs (files only, no URLs)

### LINKED ISSUE #509: swagger/openAPI does not handle remote file references
swagger allows $ref to refer to an object in another swagger file (see https://swagger.io/docs/specification/using-ref/)

This is currently not supported by the sysl importer

nofun97:
Fixes #509  .

- remote references on swagger can now be loaded into the sysl output

some thing to note:
- if two types have the same name but different field, the first one loaded stays (need an `Equal` method on the `Type` interface)
- this is assuming that in swagger v2, types are always defined in `#/definitions/`

@anz-bank/sysl-developers


nofun97:
just changing some variable and function names

PR #530, opened 2020-01-21
Added per file duplicate module import warning

### LINKED ISSUE #471: No warnings on duplicated import
### Expected behaviour
If the same module is `import`ed in the same *.sysl file more than once, a WARN should show in the output.

### Actual behaviour
When the *.sysl file `import`s the same module more than once, there's no warnings.

### Example
In the file `model_with_deps.sysl`,
```
import deps
import deps

ModelWithDeps [package="model"]:
  //...
```

nydrani:
Fixes #471.

@anz-bank/sysl-developers


Joshcarp:
Any test cases for this change?

pkg/parse/parse.go
```diff
@@ -129,10 +129,30 @@ func (p *Parser) Parse(filename string, fs afero.Fs) (*sysl.Module, error) {
 			return nil, err
 		}
 
-		antlr.NewParseTreeWalker().Walk(listener, tree)
+		localListener := NewTreeShapeListener()
```
ChloePlanet: Why do you add another locallistener instead of checking listener.imports directly after `antlr.NewParseTreeWalker().Walk(listener, tree)`?

nydrani: The 'listener' is a cumulative queue of import dependencies originating from the original module (similar to how a BFS works) so it would also contain dependencies of it's parent module etc and not reflect the dependencies of just a single module (file). So I need to create a local listener to grab just the imports of a single file.

PR #533, opened 2020-01-22
Release sysl binaries

### LINKED ISSUE #532: Docker image for go that sysl modules depends on
Currently sysl modules uses os.Execute() to get dependencies. 
This means that it's not consistent across go versions; currently It seems to work on go 12 but not go 13. 

Making a docker container would mitigate this problem as well as give it a better change of running in wasm distributions (as wasm distributions don't support the os package).

ChloePlanet:
Fixes #532  .

-
-

@anz-bank/sysl-developers

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,69 +1,81 @@
 name: Release
+
 on:
   push:
     tags:
-      - '*'
-jobs:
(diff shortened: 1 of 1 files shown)


ChloePlanet:
I'll add Dockerfile and fix #532 in another PR.

PR #542, opened 2020-01-28
fix module path

ChloePlanet:
- fix find module path by match path prefix
- fix open file outside root

@anz-bank/sysl-developers

### DIFF
--- a/pkg/mod/module.go
+++ b/pkg/mod/module.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 )
 
 var SyslModules = os.Getenv("SYSL_MODULES") != "off"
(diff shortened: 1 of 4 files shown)


pkg/mod/module.go
```diff
@@ -53,3 +53,17 @@ func Find(name string) (*Module, error) {
...
 }
+
+func hasPathPrefix(prefix, s string) bool {
+	switch {
+	default:
+		return false
+	case len(s) == len(prefix):
+		return s == prefix
+	case len(s) > len(prefix):
+		if prefix != "" && prefix[len(prefix)-1] == filepath.Separator {
```
nofun97: This part seems strange, `filepath.Clean` would remove `filepath.Separator` at the end

pkg/mod/module.go
```diff
@@ -53,3 +53,17 @@ func Find(name string) (*Module, error) {
 	}
 	return mod, nil
 }
+
+func hasPathPrefix(prefix, s string) bool {
+	switch {
+	default:
+		return false
```
nofun97: `default` should be at the end, in fact, this might be cleaner
```go
if len(s) > len(prefix){ ... }

return s == prefix
```

PR #555, opened 2020-01-30
feat: add release version to sysl info

### LINKED ISSUE #552: New Command: sysl version

ChloePlanet:
Fixes #552  .

- add new command `sysl version` to show the build info for sysl executable
- update issue template to include sysl version info
- update docs

Checklist:

@anz-bank/sysl-developers


nofun97:
Shouldn't this be part of the existing `sysl info`?

ChloePlanet:
Sure, I didn't notice the existence of `sysl info`. I'll check that.

ISSUE_TEMPLATE.md
```diff
@@ -1,7 +1,18 @@
 Please **do not post any internal, closed source snippets** on this public issue tracker!
```
cuminandpaprika: I think it might be cleaner to move the *Issue_Template* and *PullRequest_Template* into a *.github/ISSUE_TEMPLATE* folder? 
See here for an example (https://github.com/hashicorp/terraform/tree/ec90bed62d008b5c1d9ce291a03226633a7e3069/.github)

I find large root directories so intimidating!

cuminandpaprika: We can also hide the top comment using `<!--` closed by `-->` tags.

ISSUE_TEMPLATE.md
```diff
@@ -1,7 +1,18 @@
...
 
-### Expected behaviour
+### Description
 
-### Actual behaviour
+### Steps to Reproduce
+1. 
+2. 
+3.
```
cuminandpaprika: I might suggest adding something like these additional sections, so that the user has somewhere to post the logs?
```
### Debug Output
<!--
Full debug output can be obtained by running Terraform with the environment variable `TF_LOG=trace`. Please create a GitHub Gist containing the debug output. Please do _not_ paste the debug output in the issue, since debug output is long.

Debug output may contain sensitive information. Please review it before posting publicly, and if you are concerned feel free to encrypt the files using the HashiCorp security public key.
-->

### Crash Output
<!--
If the console output indicates that Terraform crashed, please share a link to a GitHub Gist containing the output of the `crash.log` file.
-->
```
I've been referencing https://github.com/hashicorp/terraform/blob/ec90bed62d008b5c1d9ce291a03226633a7e3069/.github/ISSUE_TEMPLATE/bug_report.md as an example

ISSUE_TEMPLATE.md
ChloePlanet: > I think it might be cleaner to move the _Issue_Template_ and _PullRequest_Template_ into a _.github/ISSUE_TEMPLATE_ folder?
> See here for an example (https://github.com/hashicorp/terraform/tree/ec90bed62d008b5c1d9ce291a03226633a7e3069/.github)
> 
> I find large root directories so intimidating!

I think so. I'd prefer do it in another PR. This one is for sysl binary info:)

pkg/cfg/binaryinfo.go
```diff
@@ -0,0 +1,66 @@
...
+	info, err := latestGitReleaseInfo()
+	if err != nil {
+		logrus.Errorf("Get latest git release info error: %s\n", err.Error())
+		return
+	}
+
+	Version = info[0]
+	GitCommit = info[1]
+	BuildDate = info[2]
+}
```
nofun97: Not entirely sure but probably not a good idea. Maybe some sort of a getter can be called? Maybe `GetInfo` that loads all the data and then let `Execute` call it and assigns the data. Are the global variables used anywhere else?

ISSUE_TEMPLATE.md
ChloePlanet: @cuminandpaprika I've created a pr related to this: #556

ISSUE_TEMPLATE.md
ChloePlanet: #556

PR #569, opened 2020-02-05
go mod tidy

ChloePlanet:
- go mod tidy
- add go mod tidy check to make deps
- rename check-coverage.sh
- add breaking-change tag info to pr template

Checklist:

@anz-bank/sysl-developers


nofun97:
Maybe in addition to enforcing `go mod tidy` in the Makefile, you should add a reminder somewhere if possible. Waiting for CI tests can be annoying.

ChloePlanet:
It doesn't spend more time to run `go mod tidy`, compared to previous PR checks.

Makefile
```diff
@@ -57,6 +57,13 @@ build: ## Build sysl into the ./dist folder
 
 deps: ## Download the project dependencies with `go get`
 	go get -v -t -d ./...
+	go mod tidy
+ifneq ("$(shell git status --porcelain)", "")
+	## GoReleaser has to make sure go.mod is up to date before release sysl binary. 
+	## Keep everyone remembering to update go.mod to avoid release failure.
+	echo "git is currently in a dirty state, please check in your pipeline what can be changing the following files:$(shell git diff --staged)"
+	exit 1
+endif
```
nofun97: Just remembered, if `go.mod` is tidied and there are unused dependencies, the `go.sum` will be generated again and the `git diff` will be very large. Add `go.sum linguist-generated=true` to `.gitattributes` and diff wouldn't show the difference of `go.sum`. 

Also, these lines should probably be run before `go get -v -t -d ./...`

PR #574, opened 2020-02-07
Wiring content type from swagger

### LINKED ISSUE #545: Set Content-Type for responses going out
Apparently this is used

```
func setJSONContentType(ctx context.Context) {
	headermap, _ := common.RespHeaderAndStatusFromContext(ctx)
	headermap.Add("Content-Type", "application/json")
}
```

every time the Content-Type needs to be set. Needs to be automated.

ashwinsajiv:
Fixes #545 
- Adds content type to http operation responses

@anz-bank/sysl-developers


pkg/importer/writer.go
```diff
@@ -165,6 +165,39 @@ func buildQueryString(params []Param) string {
...
+
+	contentsString = " ["
+	contents = removeContentDuplicates(contents)
+	for _, val := range contents {
+		if val.contentType == "" || val.name == "" {
+		} else {
+			contentsString = contentsString + "[" + val.contentType + ", var=\"" + val.name + "\"]"
+		}
+	}
+	contentsString = contentsString + "]"
```
golangcibot: assignOp: replace `contentsString = contentsString + "]"` with `contentsString += "]"` (from `gocritic`)

pkg/importer/writer.go
```diff
@@ -223,7 +260,7 @@ func (w *writer) writeEndpoint(method string, endpoint Endpoint) {
 	desc := getDescription(endpoint.Description)
 
 	w.writeLines(fmt.Sprintf("%s:", pathStr), PushIndent,
-		fmt.Sprintf("%s%s%s:", method, reqStr, buildQueryString(endpoint.Params.QueryParams())), PushIndent)
+		fmt.Sprintf("%s%s%s%s:", method, reqStr, responseContentType, buildQueryString(endpoint.Params.QueryParams())), PushIndent)
```
golangcibot: line is 125 characters (from `lll`)

pkg/importer/writer.go
```diff
@@ -165,6 +165,38 @@ func buildQueryString(params []Param) string {
...
+
+	contentsString := " [responses=["
+	responses = removeResponseDuplicates(responses)
+	for _, val := range responses {
+		if val.Content.contentType != "" && val.Content.name != "" {
+			contentsString = contentsString + "[" + val.Content.contentType[10:len(val.Content.contentType)-1] + "\", \"" + val.Content.name + "\"], "
+		}
+	}
+	contentsString = contentsString[:len(contentsString)-2]
+	contentsString = contentsString + "]]"
```
golangcibot: assignOp: replace `contentsString = contentsString + "]]"` with `contentsString += "]]"` (from `gocritic`)

pkg/importer/writer.go
```diff
@@ -165,6 +165,38 @@ func buildQueryString(params []Param) string {
...
+func buildResponseContentString(responses []Response) string {
+	if len(responses) == 0 {
+		return ""
+	}
+
+	contentsString := " [responses=["
+	responses = removeResponseDuplicates(responses)
+	for _, val := range responses {
+		if val.Content.contentType != "" && val.Content.name != "" {
+			contentsString = contentsString + "[" + val.Content.contentType[10:len(val.Content.contentType)-1] + "\", \"" + val.Content.name + "\"], "
```
golangcibot: line is 141 characters (from `lll`)

pkg/importer/writer.go
```diff
@@ -210,8 +243,12 @@ func buildPathString(path string, params []Param) string {
 }
 
 func (w *writer) writeEndpoint(method string, endpoint Endpoint) {
+	responseContentType := ""
```
joshcarp: nit: try using `var responseContentType string` when declaring but not initialising

pkg/importer/writer.go
```diff
@@ -210,8 +243,12 @@ func buildPathString(path string, params []Param) string {
...
 	header := buildRequestHeadersString(endpoint.Params.HeaderParams())
 	body := buildRequestBodyString(endpoint.Params.BodyParams())
+	if len(endpoint.Responses) != 0 {
+		responseContentType = buildResponseContentString(endpoint.Responses)
+	}
 	reqStr := ""
```
joshcarp: Same here (I know you didn't write this, but it's good to look out for)

pkg/importer/writer.go
```diff
@@ -165,6 +165,39 @@ func buildQueryString(params []Param) string {
...
+	contentsString += "]]"
+	if contentsString == " [responses]]" {
+		return ""
+	}
+	return contentsString
+}
+
+func removeResponseDuplicates(responses []Response) []Response {
+	keys := make(map[Content]bool)
+	list := []Response{}
```
joshcarp: Maybe try using `list := make([]Response, len(responses))`

pkg/importer/writer.go
```diff
@@ -165,6 +165,39 @@ func buildQueryString(params []Param) string {
...
+	for _, entry := range responses {
+		if _, value := keys[entry.Content]; !value {
```
joshcarp: `if _, value := keys[entry.Content]; !value {` 
This line is slightly ambiguous; usually the second value of accessing a map this way is a bool of if the key `entry.Content` is in the map, here you're calling this the `value`, even though the value would be the first value:
`if value, ok:= keys[entry.Content]; !ok` ...
If you're checking for if the duplicate is in the map then you should be using an empty struct instead of a bool

pkg/importer/writer.go
```diff
@@ -165,6 +165,39 @@ func buildQueryString(params []Param) string {
...
+	if len(responses) == 0 {
+		return ""
+	}
+
+	contentsString := " [responses=["
+	responses = removeResponseDuplicates(responses)
+	for _, val := range responses {
+		if val.Content.contentType != "" && val.Content.name != "" {
+			contentsString = contentsString + "[" +
+				val.Content.contentType[10:len(val.Content.contentType)-1] + "\", \"" + val.Content.name + "\"], "
```
joshcarp:
```contentType := val.Content.contentType[10:len(val.Content.contentType)-1]
fmt.Sprintf(`%s [ %s "," %s"]`, contentType, val.Content.name)```

File: pkg/importer/writer.go (hunk 8 above)
Comment (ashwinsajiv) [reply pkg/importer/writer.go]: 1- I care about the second variable and thus I named it value. Will change it as it creates confusion. 
2- I'm checking if !value and appending the values to a list and that is why bool is sufficient and I need to use an empty struct.

File: pkg/importer/writer.go (hunk 9 above)
Comment (ashwinsajiv) [reply pkg/importer/writer.go]: Addressed

File: pkg/importer/writer.go (hunk 5 above)
Comment (ashwinsajiv) [reply pkg/importer/writer.go]: Addressed

File: pkg/importer/writer.go (hunk 6 above)
Comment (ashwinsajiv) [reply pkg/importer/writer.go]: Addressed

File: pkg/importer/writer.go (hunk 7 above)
Comment (ashwinsajiv) [reply pkg/importer/writer.go]: Addressed

File: pkg/importer/writer.go (hunk 8 above)
Comment (ashwinsajiv) [reply pkg/importer/writer.go]: 1- I was using the bool only and that is why I chose to name it value. I see that it creates confusion so I will rename it. 

2- I am using the bool to build a map which consists of the Content as keys and the bool as value to see if it exists or not and then I append the response to the list. So I don't see why I need to use a struct?

File: pkg/importer/endpoints.go [hunk 10]
```diff
@@ -2,17 +2,20 @@ package importer
...
+	Text    string
+	Type    Type
+	Content Content
+}
+type Content struct {
+	name        string
+	contentType string
 }
 
 type Endpoint struct {
```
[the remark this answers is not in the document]
ashwinsajiv: Addressed

pkg/importer/endpoints.go
```diff
@@ -2,17 +2,20 @@ package importer
 
 // Response is either going to be freetext or a type
 type Response struct {
-	Text string
-	Type Type
+	Text    string
+	Type    Type
+	Content Content
+}
+type Content struct {
```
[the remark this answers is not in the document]
ashwinsajiv: Addressed

anzdaddy:
What's blocking this? It was approved 9 days ago.

ashwinsajiv:
After a chat with Gregg, there seems to be an issue which needs to be addressed in the importer. Hence, leaving this with the WIP tag and will merge it once its dependency is addressed.

ashwinsajiv:
> What's blocking this? It was approved 9 days ago.

After a chat with Gregg, there seems to be an issue which needs to be addressed in the importer. Hence, leaving this with the WIP tag and will merge it once its dependency is addressed.

PR #612, opened 2020-02-16
Feature/attribute example

cuminandpaprika:
- Add example for reserved attributes
- Add example for alternate attribute syntax (using @)

Checklist:
- [x] Added related tests

@anz-bank/sysl-developers


demo/examples/Attributes-Reserved/1_project.sysl
```diff
@@ -0,0 +1,39 @@
+# Sysl reserves certain attributes for sequence diagram generation.
```
joshcarp: Can this example be named Reserved-Attributes instead?

cuminandpaprika: haha sure sounds good!

PR #630, opened 2020-02-20
#625 Fix error panic: runtime error: invalid memory address or nil pointer dereference

### LINKED ISSUE #625: Sysl gen datamodel without -j results in a panic (nil pointer)
sysl purushoa$ sysl data -d -o "sdsd.svg"  output.sysl
WARN[0000] root and .sysl are undefined, . will be used instead 
panic: runtime error: invalid memory address or nil pointer dereference

ericzhang6222:
Fixes #625, this is a hot fixing as @ashwinsajiv is going to help team to adopt sysl and it is important to make sure sysl can work.

- Add nil check

Checklist:
- [Done] Added related tests
- [Done] Made corresponding changes to the documentation

@anz-bank/sysl-developers


cmd/sysl/datamodelview.go
```diff
@@ -210,6 +210,9 @@ func (v *DataModelView) drawTuple(
 				collectionString = fmt.Sprintf("+ %s : **Set <%s>**\n", attrName, path[0])
 				relation = `0..*`
 			default:
+				if attrType.GetTypeRef() == nil {
```
camh-anz: I would change this to `case attrType.GetTypeRef() != nil:` and have the `default:` case just `continue`. This way it follows the same form as the rest of the switch statement.

ericzhang6222: It is a hot fix as it is blocking team's work, and this change is a temp solution, will check root cause why attrType.GetTypeRef() return nil. So now it is a simple change.

camh-anz: I'm sorry, but the change I am requesting is very simple - it is fewer lines of code than what you have done. Please keep the code clean.
```
case attrType.GetTypeRef() != nil:
	path = attrType.GetTypeRef().GetRef().Path
	collectionString = fmt.Sprintf("+ %s : **%s**\n", attrName, path[0])
	relation = `1..1 `
default:
	continue
```

ericzhang6222: Fixed.

PR #642, opened 2020-02-25
Issue 619 - update code to simplify codegen command

### LINKED ISSUE #619: Simplify sysl codegen command
### Current Sysl Version
`sysl codegen --transform=one.sysl --grammar=grammar.go.gen --start=goFile --out=gen/ --app-name=foo model.sysl`
This currently needs to be repeated for every transform for every app

A better approach would be to use a config issue or something:
`sysl codegen --config=config.yaml --app-name=simple out=gen/ model.sysl`

Which can then infer all the `server` and `client` (possibly by putting the required transforms in server/ and client/ dirs) and having the required `grammar.go.gen` somewhere

### Attempted Solutions
Currently a long and tedious `makefile` is used
EDIT:
I don't like the idea of an environment variable

ericzhang6222:
Fixes #619  .

- Add config flag to place some flags of codegen command

Checklist:
- [x]  Added related tests
- [x]  Made corresponding changes to the documentation


cmd/sysl/cmd_codegen.go
```diff
@@ -70,3 +84,26 @@ func (p *codegenCmd) Execute(args ExecuteArgs) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to read config file %s", p.config)
+		}
+
+		p.transform = syslutil.ResetVal(p.transform, config.Transform)
+		p.grammar = syslutil.ResetVal(p.grammar, config.Grammar)
+		p.depPath = syslutil.ResetVal(p.depPath, config.DepPath)
+		p.basePath = syslutil.ResetVal(p.basePath, config.BasePath)
+		p.appName = syslutil.ResetVal(p.appName, config.AppName)
+		p.start = syslutil.ResetVal(p.start, config.Start)
```
golangcibot: config.Start undefined (type config.CmdContextParamCodegen has no field or method Start) (from `typecheck`)

cmd/sysl/cmd_codegen.go
```diff
@@ -70,3 +84,25 @@ func (p *codegenCmd) Execute(args ExecuteArgs) error {
 	}
 	return outputToFiles(output, syslutil.NewChrootFs(args.Filesystem, p.outDir))
 }
+
+func (p *codegenCmd) loadFlags() error {
+	err := validate.CodeggenRequiredFlags(p.config, p.grammar, p.transform)
```
anz-gordonj7: 1) typo in CodegenRequiredFlags
2) Why did you choose too put that function in the validate package? It is only used by this command, and is rightly coupled to this command, it really shuold just be here.

pkg/syslutil/setValue.go
```diff
@@ -0,0 +1,9 @@
+package syslutil
+
+func ResetVal(init string, new string) string {
```
anz-gordonj7: I dont understand what this function is supposed to be used for?

It does not reset any values, it returns the new value if and only iff the init value is empty. Function naming is very important.

pkg/config/cmd_config.go
```diff
@@ -0,0 +1,29 @@
+package config
+
+import (
+	"io/ioutil"
+
+	"github.com/ghodss/yaml"
+)
+
+type CmdContextParamCodegen struct {
+	Transform string
```
anz-gordonj7: this struct should have yaml tags

And this config should be with the codegen code, not in a seperate package.

pkg/config/cmd_config.go
```diff
@@ -0,0 +1,29 @@
+package config
+
+import (
+	"io/ioutil"
+
+	"github.com/ghodss/yaml"
```
anz-gordonj7: we already use the `gopkg.in/yaml.v2` yaml package, no need to introduce another dependancy

cmd/sysl/cmd_codegen.go
ericzhang6222: This method is added to check 

> As mentioned, I don't see the value-proposition of this change. You're simply moving command arguments from the command line to a seperate file which does very little as far as user experience is concerned.
> 
> Now if thats all we want to do then I have a much better solution. gcc (and other programs) support a `@file` command line option which just reads the contents of the file as the command line arguments (it is used because gcc can get REALLY long command lines with dozens of flags). How about simply supporting that and using it for all the sysl commands as an option?
> 
> see the very bottom of https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html#Overall-Options ( @file)
> 
> how you'd d this:
> in `cmd/sysl/sysl.go main3()` you'd scan the supplied args list, for any arg which looks like `@filename` , the filename would be looked up, if it exists it gets read into the args in the current position replacing the @filename flag, _then_ once they are all consumed pass the new command line args to kingpin for processing. none of the actual commands need to be changed to support this.

@anz-gordonj7 , this change is a simple fixing to make sure it can be delivered soon and @joshcarp can use it in next week's show case. And after this is done, I will change more on this piece, I think then I can try your solution. Thoughts?

ericzhang6222: validate package is setup from validation purpose, and method CodeggenRequiredFlags is used to do validation too, so I add it here to reduce code in cmd_codgen.go.

pkg/config/cmd_config.go
ericzhang6222: As Josh has updated code and moved types to package cmdutils, I removed CmdContextParamCodegen here and use the one in cmdutils, also added yaml tags.

pkg/config/cmd_config.go
ericzhang6222: I don't want to introduce new dependency, but actually this package "github.com/ghodss/yaml" has been used in code if you check, so I think this one is an option too. And I have changed it to "gopkg.in/yaml.v2".

cmd/sysl/cmd_codegen.go
anz-gordonj7: > @anz-gordonj7 , this change is a simple fixing to make sure it can be delivered soon and @joshcarp can use it in next week's show case. And after this is done, I will change more on this piece, I think then I can try your solution. Thoughts?

I honestly never want to hear the phrase "is a simple fixing to make sure it can be delivered soon". nothing is stopping @joshcarp from using your branch in a demo (or some other hack/fix), doesnt mean it needs to be in master. I don't want to see half-patches merged thank you.

> validate package is setup from validation purpose, and method CodeggenRequiredFlags is used to do validation too

It is *supposed* to be validating a sysl file, not the command line options. This is absolutly the wrong place to put this code.

pkg/syslutil/setValue.go
ericzhang6222: Yes, my bad, has changed name to GetNoneEmpty.

cmd/sysl/cmd_codegen.go
ericzhang6222: @anz-gordonj7 , thank you for your update and you are right. But simple fixing doesn't mean fixing with bad implementation or code, right? I raise PR because I am considering team will use this feature after @joshcarp's showcase. And anyway I have pinged @joshcarp and see if he can use my branch to do showcase.

ericzhang6222: @anz-gordonj7, and in the same time,  I am trying your suggestion :
how you'd d this:
in cmd/sysl/sysl.go main3() you'd scan the supplied args list, for any arg which looks like @filename , the filename would be looked up, if it exists it gets read into the args in the current position replacing the @filename flag, then once they are all consumed pass the new command line args to kingpin for processing. none of the actual commands need to be changed to support this.

ericzhang6222: Fixed in another branch issue-619-new-fixing

ericzhang6222:
This approach is abandoned.

PR #643, opened 2020-02-25
Update docs for dockerfile

cuminandpaprika:
- Fix broken link 
- Update instructions for using sysl with the docker image

Checklist:
- [x] Added related tests

@anz-bank/sysl-developers

### DIFF
--- a/docs/website/content/docs/installation.md
+++ b/docs/website/content/docs/installation.md
@@ -57,7 +57,7 @@ $ docker run --rm \
   -w /go/src/github.com/anz-bank/sysl \
   anzbank/sysl:latest validate -v ./demo/examples/Modules/model_with_deps.sysl
 ```
-We have used this [Dockerfile](Dockerfile) to create this image.
+We have used this [Dockerfile](https://github.com/anz-bank/sysl/blob/master/Dockerfile) to create this image.
 
 
 ## Compiling from source
--- a/docs/website/content/docs/installation.md
+++ b/docs/website/content/docs/installation.md
@@ -45,13 +45,24 @@ $ sysl help
 
 ## Running with Docker
 
-You can also use it within a [Docker container](https://hub.docker.com/r/anzbank/sysl). To do that, you’ll need to execute something more-or-less like the following:
+You can also use it within a [Docker container](https://hub.docker.com/r/anzbank/sysl).
+
+First get the docker image using
 
 ```
-$ docker run --rm anzbank/sysl:latest help
+$ docker pull anzbank/sysl:latest
 ```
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -3,7 +3,7 @@
 #
 # The produced image is published to https://hub.docker.com/r/anzbank/sysl
 
-FROM golang:1.13-alpine
```
ChloePlanet: The reason to use `golang:1.13-alpine` is `go mod` commands which sysl modules is based on require installed `go`.

docs/website/content/docs/installation.md
```diff
@@ -45,19 +45,30 @@ $ sysl help
...
 ```
 
+For MacOS and Linux Users
+
+```bash
+$ alias sysl="docker run --rm -it -v $HOME:$HOME -w $(pwd) anzbank/sysl:latest"
+$ sysl info
 ```
+
+Sysl can then be used from the same terminal window subsequently. Alternatively, it can also be added to the `.bashrc` file to add the `sysl` command permanently.
```
ChloePlanet: ...`.bashrc` or `.zshrc` file...

Dockerfile
cuminandpaprika: Oh wait, do some of the sysl codegen commands use go mod internally?

ChloePlanet: Yes, if it imports sysl files from external repo.

docs/website/content/docs/installation.md
cuminandpaprika: Awesome, will fix this up!

Dockerfile
cuminandpaprika: From my understanding, the build process occurs in a different environment as part of the goreleaser workflow, this dockerfile just copies over the binary without running a build?

So, that means that we actually don't need the `golang:1.13-alpine` image, we can just use alpine

Happy to talk about this in person, if I've got the wrong understanding of the dependencies here!

ChloePlanet: > From my understanding, the build process occurs in a different environment as part of the goreleaser workflow, this dockerfile just copies over the binary without running a build?

Yes you are right about this. 
As for the base image, if we use `alpine`, the sysl binary will raise "command not found: go" error when it runs this function https://github.com/anz-bank/sysl/blob/master/pkg/mod/gomod.go#L80. Hence `golang:1.13-alpine` is required.

cuminandpaprika: Okay turns out Go Get is used by the Sysl Modules functionality so we should stick with `golang:1.13-alpine`, will remove the changes!

Issue #640, opened 2020-02-25
Sysl AST printer


joshcarp:
It would be nice to print the sysl ast out into the original source code again. 

This would be good for generating sysl code from external sources (converting .proto files to .sysl files) and could simplify importers by having ony one data structure (currently there are a few in the importer package)

This could also be used to make a sysl formatter kinda like `go fmt`

PR #653, opened 2020-02-26
Sysl printer

joshcarp:
Starts on #640 

- Add printer package that can print out sysl data models back into source code

This doesn't reach full fidelity, for example
- `[attribute= "foo"]` gets turned into `@attribute = "foo"`
- Order of applications and endpoints (anything stored into a map) is printed out alphabetically
- Doesn't yet support more complex sysl files like `Projects` or the weird :: syntax 

- REST syntax like:
```
  /moredep:
    /{id<:int}:
      GET:
        return Dep
      POST:
        return Dep
```

gets turned into (still valid)

```
GET /moredep/{id<:int}:
    return Dep
POST /moredep/{id<:int}:
    return Dep
```

Checklist:
- [x] Added related tests

@anz-bank/sysl-developers

### DIFF
--- a/pkg/printer/alphabetical.go
+++ b/pkg/printer/alphabetical.go
@@ -1,0 +1,49 @@
+// alphabetical.go are the functions to get a sorted list of keys from sysls map types
+// Unfortunately go doesn't allow generics and pointer types don't implement empty interfaces
+// so map[string]interface{} doesn't cut it
+
+package printer
+
+import (
+	"sort"
+
+	"github.com/anz-bank/sysl/pkg/sysl"
+)
+
+func alphabeticalAttributes(m map[string]*sysl.Attribute) []string {
+	keys := make([]string, 0, len(m))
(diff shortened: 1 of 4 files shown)


pkg/printer/alphabetical.go
```diff
@@ -0,0 +1,49 @@
+// alphabetical.go are the functions to get a sorted list of keys from sysls map types
+// Unfortunately go doesn't allow generics and pointer types don't implement empty interfaces
+// so map[string]interface{} doesn't cut it
```
anz-gordonj7: you could probably use `map[string]proto.Message` for this no?

joshcarp: `Cannot use 'a.Attrs' (type map[string]*Attribute) as type map[string]*proto.Message`

because it's a map with the value being a pointer to a type the `proto.Message` thing doesn't work because a pointer

PR #655, opened 2020-02-28
First test script to install sysl using homebrew #minor

ashwinsajiv:
Adding a test script to the sysl repo to install sysl using homebrew. Does not impact any other file.

### DIFF
--- a/test.rb
+++ b/test.rb
@@ -1,0 +1,36 @@
+require 'rubygems/package'
+require 'zlib'
+
+TAR_LONGLINK = '././@LongLink'
+tar_gz_archive = File.join(ENV['HOME'], 'Library', 'Caches', 'Homebrew', 'sysl--0.6.3.tar.gz')
+destination = File.join(ENV['HOME'], 'Library', 'Caches', 'Homebrew')
+
+Gem::Package::TarReader.new( Zlib::GzipReader.open tar_gz_archive ) do |tar|
+  dest = nil
+  tar.each do |entry|
+    if entry.full_name == TAR_LONGLINK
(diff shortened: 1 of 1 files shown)


cuminandpaprika:
Got a question, what's the reason that the homebrew script needs to live in the same repo as Sysl? 

Okay, post our discussion, sounds like the ruby script is building and copying the file to the gopath, which we're currently not doing in the makefile.

ashwinsajiv:
> Got a question, what's the reason that the homebrew script needs to live in the same repo as Sysl?
> 
> Okay, post our discussion, sounds like the ruby script is building and copying the file to the gopath, which we're currently not doing in the makefile.

Yes, also the make file install has access to sysl binary already through the dist folder whereas the homebrew installation has the zip file of the release which does not have the binary. Hence, the decision to have a seperate script to do this. The script is also intended to do more - such as -  install go and create gopath if it doesn't exist already.

test.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+    dest = nil
+  end
+end
+
+Src = File.join(ENV['HOME'], 'Library', 'Caches', 'Homebrew', 'sysl-0.6.3', 'cmd', 'sysl')
+Dir.chdir Src
+system("go", "build")
+Dest = File.join(ENV['HOME'],'go', 'bin')
+FileUtils.cp(File.join(Src, 'sysl') ,Dest)
```
camh-anz: Missing eol at eof

ashwinsajiv: I have an issue raised which keeps track to rename the file and to clean things up.

PR #658, opened 2020-02-28
add escaping to endpoints

### LINKED ISSUE #624: Allow method to escape characters in endpoint definitions
```
fakerock[package="fakerock"]:
    /.foo:
            GET:
                return string
```

the . in the endpoint breaks sysl, with the error 

```
line 8:5 mismatched input '.' expecting {COLON, SQ_OPEN}
line 9:12 mismatched input 'GET' expecting {'!wrap', '!table', '!type', '!alias', '!union', '!view', WHATEVER, MIXIN, '<->', '.. * <- *', FORWARD_SLASH, '@', SYSL_COMMENT, TEXT_LINE, Name, E_Name}
line
```

there should be a way to escape these characters

nofun97:
Fixes #624  .

- Endpoints are now unescaped which means you can escape special characters
- To escape characters, please use an URL encoded characters. Which means, if you would like to create `test.stuff` endpoints, you would type `test%2e.stuff` in your sysl file.
- Works with exporter, importer, and validator.

Checklist:
- [x] Added related tests

@anz-bank/sysl-developers


pkg/importer/utils.go
```diff
@@ -67,3 +70,10 @@ func spaceSeparate(items ...string) string {
 	}
 	return strings.Join(t, " ")
 }
+
+func getSyslSafeEndpoint(endpoint string) string {
+	// url.PathEscape does not escape '.'
+	escaped := strings.ToLower(url.PathEscape(endpoint))
```
golangcibot: ineffectual assignment to `escaped` (from `ineffassign`)

pkg/grammar/SyslLexer.g4
```diff
@@ -191,7 +191,7 @@ TEXT_LINE       :
                 { !startsWithKeyword(ls(p), p.GetText()) }?
                 ;
 
-Name            : [a-zA-Z_][-a-zA-Z0-9_]*;
+Name            : [%a-zA-Z_][-%a-zA-Z0-9_]*;
```
anz-gordonj7: this regex isnt actually correct for url escaping, but its close enough I guess.... It really should be something like `(%[0-9a-fA-F][0-9a-fA-F])?`

pkg/importer/utils.go
anz-gordonj7: uppercase hex is usually more common and a bit easier to read.

pkg/grammar/SyslLexer.g4
nofun97: that would enforce it better, but what would the regex for `Name` be?

nofun97:
don't merge this yet, I just found some more bugs.

PR #663, opened 2020-03-02
Fix case which flag app-name has double quotes and add examples

### LINKED ISSUE #619: Simplify sysl codegen command

ericzhang6222:
Fixes #619 .

- Fix case which flag app-name has double quotes and add examples

Checklist:
- [x] Added related tests


pkg/syslutil/cmd_config.go
```diff
@@ -12,8 +13,12 @@ func ReadCMDFlags(configPath string) ([]string, error) {
 	if ferr != nil {
 		return flags, ferr
 	}
-	str := string(data)
-	flags = strings.Fields(strings.TrimSpace(str))
+	re := regexp.MustCompile("(\\S+\"[^\"]+\")|\\S+")
```
anzdaddy:
```suggestion
	re := regexp.MustCompile(`(\S+"[^"]+")|\S+`)
```

pkg/syslutil/cmd_config.go
```diff
@@ -12,8 +13,12 @@ func ReadCMDFlags(configPath string) ([]string, error) {
...
+	flags = re.FindAllString(strings.TrimSpace(string(data)), -1)
```
anzdaddy: Initialise `flags` here.

pkg/syslutil/cmd_config.go
```diff
@@ -12,8 +13,12 @@ func ReadCMDFlags(configPath string) ([]string, error) {
...
+
+	for i, flag := range flags {
+		flags[i] = strings.ReplaceAll(flag, "\"", "")
```
anzdaddy: This is dangerous. Sometimes you want to pass a double-quote as part of a parameter. Better to use a regex to match a common flag-or-arg pattern: `^(-+[^=]+=)?"?(.*?)"?$`.

pkg/syslutil/cmd_config.go
ericzhang6222: Fixed.

pkg/syslutil/cmd_config.go
ericzhang6222: Fixed.

pkg/syslutil/cmd_config.go
ericzhang6222: Checking

ericzhang6222: Add a new pattern to parse flags config: `(-+[^=\s\n]+)?((=?"[^=\n]+")|(=?[^=\s\n]+))?\s*`

PR #667, opened 2020-03-04
Fix Docker file

### LINKED ISSUE #666: Docker image will not run on Docker desktop
### Description

When running the Docker image with

```
docker container run --rm anzbank/sysl:latest help`
```

an error occurs

```
standard_init_linux.go:211: exec user process caused "no such file or directory"
```

### Steps to Reproduce
1. run the command `docker container run --rm anzbank/sysl:latest help` at the PowerShell of WSL prompt 

**Expected behavior:**

Some nice help text is displayed

**Actual behavior:**
```
standard_init_linux.go:211: exec user process caused "no such file or directory"
```

* Debug&Crash Output

N/A

Client:
 Version:           19.03.3
 API version:       1.40
 Go version:        go1.12.10
 Git commit:        2355349d-
 Built:             10/14/2019 16:41:26
 OS/Arch:           windows/amd64
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          19.03.5
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.12
  Git commit:       633a0ea
  Built:            Wed Nov 13 07:29:19 2019
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          v1.2.10
  GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
 runc:
  Version:          1.0.0-rc8+dev
  GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683

Windows 10 Pro

alecthegeek:
Fixes #666 .

1. Move Dockerfile into its own directory -- smaller build context -> faster build
2. Use Debian Buster instead of Alpine. Fixes previous startup problems
3. Build sysl inside container instead of copying unknown version
4. Using Debian makes it easier to build on top of this image using `apt`

Buld image with

```
 docker image build -t anzbank/sysl:<tag> docker-build-context
```

Notes: This version defaults to Go 1.4 and Sysl 0.7.1. Override version
defaults with `build-args`.

My image size is  96.9MB, which seems OK to me. The current official image is 379MB but it does not work.

Checklist:
- [ ] Added related tests  -- N/A
- [ ] Made corresponding changes to the documentation  N/A

@anz-bank/sysl-developers


camh-anz:
> Use Debian Buster instead of Alpine. Fixes previous startup problems

What problems are these? Is this the issue described in #666? If so, this is not a good "fix" - a Dockerfile should never depend on the outside build environment, so even if this PR would make that work, lets not document that.

> Using Debian makes it easier to build on top of this image using apt

You could say that it should instead use a redhat image as that would be easier to build on top of using yum/rpm. My point here is that introducing Debian into the image for personal reasons should not be done. It needs a good justification. And keeping a small tight image (alpine) is generally preferable that a larger one. A user who wants a Debian-based image could easily build their own.

Some other comments incoming on the specific of the commit...

alecthegeek:
> You could say that it should instead use a redhat image as that would be easier to build on top of using yum/rpm. My point here is that introducing Debian into the image for personal reasons should not be done. It needs a good justification. And keeping a small tight image (alpine) is generally preferable that a larger one. A user who wants a Debian-based image could easily build their own.

This is not a "personal preference".

1. The apt package manager is very well understood, apk is not. It is likely that Sysl image will be tailored by users with additional tools for their own environment. Apt makes this easier.

2. Alpine has had security issues and many people are nervous about using it.

> What problems are these? Is this the issue described in #666? If so, this is not a good "fix" - a Dockerfile should never depend on the outside build environment, so even if this PR would make that work, lets not document that.


The problem you described in #666. The Alpine image will need additional libs to work. Debian does not.

alecthegeek:
Just notice I skipped pandoc install.

docker-build-context/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# RUN find / type f -name sysl -ls
+
+FROM debian:buster-slim as runner
+
+COPY --from=builder  /go/bin/sysl /usr/local/bin/
+
+ENTRYPOINT ["/usr/local/bin/sysl"]
+
+CMD ["help"]
```
camh-anz: Missing end-of-line at end-of-file. Every line should end with a newline.

docker-build-context/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# Goreleaser builds a docker image according to this Dockerfile by copying the sysl binary 
+# to a golang:1.13-alpine image and setting up the entrypoint.
+#
+# The produced image is published to https://hub.docker.com/r/anzbank/sysl
+ARG go_ver=1.14
```
camh-anz: The current `go.mod` has 1.13. Bumping to 1.14 should be fine, but should not be part of this PR.

docker-build-context/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+WORKDIR /work
+
+RUN git clone --branch ${sysl_ver} --single-branch --depth 1 https://github.com/anz-bank/sysl.git && \
+    cd sysl && \
+    go get ./... && \
+    go install ./cmd/sysl
+
+# RUN find / type f -name sysl -ls
+
+FROM debian:buster-slim as runner
```
camh-anz: For a release, the FROM image should be a specific one that will not change. Otherwise the image build is not reproducible.

docker-build-context/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+COPY --from=builder  /go/bin/sysl /usr/local/bin/
+
+ENTRYPOINT ["/usr/local/bin/sysl"]
```
camh-anz: I'm not sure if we had an image that ever worked. If we did, moving the binary from `/sysl` to `/usr/local/bin/sysl` might break some users. Furthermore, `/usr/local/bin` make little sense inside a container, when a container is meant to be a single process, not an entire operating system. Short paths are good paths I think.

docker-build-context/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+# The produced image is published to https://hub.docker.com/r/anzbank/sysl
+ARG go_ver=1.14
+
+FROM golang:${go_ver}-buster as builder
+
+ARG sysl_ver=v0.7.1
+
+WORKDIR /work
+
+RUN git clone --branch ${sysl_ver} --single-branch --depth 1 https://github.com/anz-bank/sysl.git && \
```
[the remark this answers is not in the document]
alecthegeek: I think you are assuming this will be used in the context of a checked out repo mounted via a bind mount?

That would not allow auto builds to occur on Docker Hub. If it works differently for you let me know.

docker-build-context/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+    cd sysl && \
+    go get ./... && \
+    go install ./cmd/sysl
```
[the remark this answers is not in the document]
alecthegeek: On purpose? I thought since 1.11 we are no longer meant to build in GOPATH.

But I don't do much go so ready to be corrected.

docker-build-context/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# RUN find / type f -name sysl -ls
```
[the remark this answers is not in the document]
alecthegeek: Yes -- I removed that about 10m ago :-(

camh-anz:
> The Alpine image will need additional libs to work. Debian does not.

Additional libs are not needed - it just needs to be built in an alpine container if it going to be deployed in an alpine container. For example, the following works in alpine (after removing `pkg` from `.dockerignore`):
```
FROM golang:1.13.8-alpine3.11 as builder

WORKDIR /src
COPY . .
RUN go install ./...

FROM alpine:3.11.3
COPY --from=builder /go/bin/sysl /sysl
ENTRYPOINT ["/sysl"]
CMD ["help"]
```

docker-build-context/Dockerfile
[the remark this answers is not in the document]
alecthegeek: Personal preference. I prefer to document the check out in the Dockerfile rather than relying on having the correct version lying around in the build context.

Will follow the product conventions.

camh-anz:
> This is not a "personal preference".

Sorry, I didn't mean to imply anything in particular here, just that with no justification for a change, I can only assume a personal preference.

If you would like to change the base image for the sysl container image from alpine to debian, that can be discussed, but should be done independently of getting the Dockerfile working.

> Alpine has had security issues and many people are nervous about using it.

Sorry if this is too blunt, but this comment is just too hand-wavey. If there are specific issues, they should be pointed out. Alpine is often chosen as a base as it presents a smaller attack surface than a larger distro like Debian. If alpine has security issues, I'd be more inclined to go distroless (https://github.com/GoogleContainerTools/distroless) than add a fat base.

> It is likely that Sysl image will be tailored by users with additional tools for their own environment.

I'd like to understand this more before accepting it. The sysl container image is there to make it easy to run the sysl binary. It has defined inputs and outputs. Any pre-processing of the input or post-processing of the output does not need to be done in the sysl container image.

This is not to say there may not be use cases for a broader image, but I'd like to understand them first before making a change to move in that direction.

docker-build-context/Dockerfile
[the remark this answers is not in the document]
alecthegeek: Well when I copy the source in via a COPY it does not build. The build error is

```
build github.com/anz-bank/sysl/cmd/sysl: cannot load github.com/anz-bank/sysl/pkg/cfg: no matching versions for query "latest"
```

(This is using a fresh clone at the v0.7.1 tag)

Cloning the same tag into the image and building works fine :-(.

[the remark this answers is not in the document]
camh-anz: Yeah, the `.dockerignore` is broken. You'll need to remove `pkg` from it (and may as well remove the `!pkg/sysl_js` line too since that's now useless).

I added `pkg/antlr-4.7-complete.jar` and `scripts/generate` to it since they aren't needed for the docker build and just bloat the context.

camh-anz:
> So I've been working with the Alpine image and getting Plantuml installed (well Java really) is a PITA. I don't want to go through that

Good. I don't think it's really necessary. The container image should just contain sysl, IMHO. PlantUML is different software that belongs in its own image. I usually run the sysl tests using `plantuml/plantuml-server:tomcat` from dockerhub when I don't want to use the public `www.plantuml.com` url.

alecthegeek:
Still using debian base image because it's too hard to install plantuml in Alpine

Still using Git Checkout because

1. It seems better practice to build directly from a know tag instead of what could be lying around in the build context
2. I can't build successfully from the build context on my machine -- be great if someone else could test

camh-anz:
> Still using debian base image because it's too hard to install plantuml in Alpine

Please don't put plantuml in a sysl image. plantuml belongs in its own image. Containers are not VMs - you don't load them up with everything you want. A container is a wrapper around a process. sysl and plantuml are separate processes and belong in separate containers. Furthermore, sysl runs as a job - it runs and exits. Plantuml runs as a server. It keeps running until you tell it to exit. I can't see how these two programs are meant to co-exist in the same image.

> It seems better practice to build directly from a know tag instead of what could be lying around in the build context

The build from CI which is pushed to dockerhub does not have anything "lying around". It is a fresh checkout of the code at the point the release is made.

If you are developing locally, you want your code that you have not yet pushed upstream to be buildable in a container, otherwise it is not possible to test that before merging. It NEEDS to be built from your current workspace, not the git repo.

Finally, as I mentioned earlier, since CI automatically tags and builds a release, the docker build will always build an old release with the method you are using. It it needs to build the code just tagged, but you will still have the old version in the dockerfile.

Standard practice is to do a dockerbuild on whatever you have in your workspace. That is the point of the docker context. By passing an empty context and pulling the code into the docker image is bypassing the docker context. It is there for a reason, please use it.

> I can't build successfully from the build context on my machine -- be great if someone else could test

You need make the changes to the `.dockerignore` that I said would be needed - i.e remove `pkg` (and `!pkg/sysl_js`) - it was in reply to one of your comments, so it was easy to miss further up the page. With those changes and the dockerfile I added to the comments above, I was able to build a image that could run `sysl help`.

However, my colleagues tell me that sysl will need to have Go installed as it makes use of the Go module system. So the base image for the release will need to be a golang image. I'm not too happy about that, but that's the way it is now.

alecthegeek:
I think we are there?

.dockerignore
```diff
@@ -2,3 +2,7 @@
 pkg
 !pkg/sysl_js
 **/node_modules
+/dist/
+/release/
+/pkg/antlr*.jar
```
camh-anz: prefer to keep lists of files, etc in alphabetical order

.dockerignore
```diff
@@ -2,3 +2,7 @@
 pkg
```
camh-anz: this needs to be removed as most of the sysl code is in there. It wont build without this directory.

.dockerignore
```diff
@@ -2,3 +2,7 @@
...
 !pkg/sysl_js
```
camh-anz: This could be removed onced `pkg` is removed, as this line becomes useless.

.dockerignore
```diff
@@ -2,3 +2,7 @@
...
 **/node_modules
+/dist/
+/release/
+/pkg/antlr*.jar
+/scrpts/generate/
```
camh-anz: `scrpts` is misspelt. `generate` is a file not a directory, so it should not have a trailing slash.

Dockerfile
```diff
@@ -1,10 +1,34 @@
...
+
+RUN test -n "${sysl_tag}" || ( echo Plese set sysl_tag build-arg && exit 1 )
 
-FROM golang:1.13-alpine
+RUN apk --no-cache add git
 
-COPY sysl /
+WORKDIR /sysl
+
+RUN git clone --branch ${sysl_tag} --single-branch --depth 1 https://github.com/anz-bank/sysl.git . && \
```
camh-anz: I don't know what else to say. I feel this is 100% wrong. I want `docker build .` to build what is in `.` just like every other docker-build-from-source dockerfile I have does. I need to be able to docker build my local source code before I raise a PR so I can verify it works properly in a docker image. Being only able to build a docker image of the official source code for which there is a git ref (branch/tag) is just too constraining.

Dockerfile
```diff
@@ -1,10 +1,34 @@
...
+    go install ./cmd/sysl
+
+FROM alpine:${alpine_ver} as runner
```
camh-anz: "as runner" is unnecessary here. please remove.

As mentioned in one of my comments, `sysl` uses the `go` command for modules, so the image needs to be a `golang` one.

Using 3.11 as `${alpine_ver}` on the final image means that this is not a reproducible build - the 3.11 tag moves as additional 3.11 point releases are made. I prefer to use fixed tags for final images so that the image can be reproducibly built from the same commit.

Dockerfile
```diff
@@ -1,10 +1,34 @@
...
+
+COPY --from=builder /go/bin/sysl /
+
+WORKDIR /work
```
camh-anz: What is the purpose of this? It seems unnecessary and arbitrary.

Dockerfile
```diff
@@ -1,10 +1,34 @@
...
+#    docker image build --build-arg sysl_tag=v0.7.1  -t my-sysl:0.7.1 .
+
+ARG go_ver=1.13
+ARG alpine_ver=3.11
+
+FROM golang:${go_ver}-alpine${alpine_ver} as builder
+
+ARG sysl_tag
+
+RUN test -n "${sysl_tag}" || ( echo Plese set sysl_tag build-arg && exit 1 )
```
camh-anz: "Plese" is misspelt.

Dockerfile
```diff
@@ -1,10 +1,34 @@
...
 
-FROM golang:1.13-alpine
+RUN apk --no-cache add git
 
-COPY sysl /
+WORKDIR /sysl
+
+RUN git clone --branch ${sysl_tag} --single-branch --depth 1 https://github.com/anz-bank/sysl.git . && \
+    go install ./cmd/sysl
```
camh-anz: No need to keep this on the same command line as the `git clone`. Make it a separate `RUN` statement. This is an intermediate build image and does not make it into the final image, so we don't need to contort command lines to accommodate docker layering here.

camh-anz: One more thing - `sysl` needs to be built with `make build` so that the build tags are set in the binary so that `sysl info` will output the correct information.

Dockerfile
alecthegeek: Now N/A

Dockerfile
alecthegeek: Now N/A

Dockerfile
alecthegeek: Personal habit

Now removed

Dockerfile
alecthegeek: I use "as runner" to give the reader extra hints if ever extra layers are added. It seems like good hygiene and does no harm

Changed to golang

I prefer my images to pick up fixes so I don't usually provide the patch version. i.e. if Go get some fixes I want them the next time I build.

However I have changed to specify the full version as requested

camh-anz: Yeah, I see both sides of this, but there are certain environments where reproducible builds are a requirement. We should consider using tools like renovate (https://github.com/renovatebot/renovate) to automatically bump the version so we get the best of both worlds.

Dockerfile
```diff
@@ -1,10 +1,26 @@
...
 # The produced image is published to https://hub.docker.com/r/anzbank/sysl
 
-FROM golang:1.13-alpine
+ARG go_ver=1.13.8
+ARG alpine_ver=3.11
 
-COPY sysl /
+FROM golang:${go_ver}-alpine${alpine_ver} as builder
+
+RUN apk --no-cache add make
```
camh-anz: I think you're going to need to keep git here. `make build` depends on it for setting the build version info. When I build your image, I get:
```
$ docker run --rm sysl info
Build:
  Version      : --2020-03-05T21:00:17Z
  Git Commit   :
  Date         : 2020-03-05T21:00:17Z
  Go Version   : unspecified
  OS           : unspecified
Runtime:
  GOOS/GOARCH  : linux/amd64
  Go Version   : go1.13.8
```

The version is missing the short sha and branch name, and the git commit is missing the short sha. I wonder if it might be best to pass them in from outside, but for now, let's just get it going with what's there and maybe refactor it later.

`make build` does not set the build Go Version or OS, but that's the way it is. I'll raise a follow-up PR to fix that.

alecthegeek:
```
Build:
  Version      : a20c98f-docker-build-2020-03-06T01:03:12Z
  Git Commit   : a20c98f
  Date         : 2020-03-06T01:03:12Z
  Go Version   : unspecified
  OS           : unspecified
Runtime:
  GOOS/GOARCH  : linux/amd64
  Go Version   : go1.13.8
```

That seems to be fixed

camh-anz:
If you can just fix those two outstanding issues, I'll merge this PR.

alecthegeek:
> If this PR were to be merged as is, the goreleaser machinery would break because when it builds the dockerfile it does not pass the build arg for the sysl tag. While I strongly believe having the docker build only be able to build official source from existing refs is wrong, this PR is still incomplete without updating the goreleaser config.

Not sure what you need here? Is this fixed by using `make` to build?

camh-anz:
> Not sure what you need here? Is this fixed by using `make` to build?

You can ignore that. It was an out-of-date comment based I think on my review crossing with your commit. It was referencing that the docker build required an argument and would fail without it. The goreleaser configs were not providing that argument. But it is moot now that the argument is not required.

alecthegeek:
K -- so we are done? I think I've fixed the rest

camh-anz:
> K -- so we are done? I think I've fixed the rest

Still waiting for those two gitignore changes.

alecthegeek:
Already done AFIK

```
**/node_modules
/dist/
/pkg/antlr*.jar
/release/
/scripts/generate
```

camh-anz:
> Already done AFIK

That's odd. I don't see that change. When I look at the "Files Changed" tab, I still see the original content, and github is not marking the review comments as "Outdated". Can you double-check you've pushed your latest commit?

alecthegeek:
> > Already done AFIK
> 
> That's odd. I don't see that change. When I look at the "Files Changed" tab, I still see the original content, and github is not marking the review comments as "Outdated". Can you double-check you've pushed your latest commit?

Looks good on GitHub?

https://github.com/alecthegeek/sysl/blob/docker-build/.dockerignore

camh-anz:
> Looks good on GitHub?
> 
> https://github.com/alecthegeek/sysl/blob/docker-build/.dockerignore

Screen Shot 2020-03-22 at 20 09 47

This is what I see. And I see it from two separate computers. What do you see?

alecthegeek:
Something went squiffy.

I pulled merged and pushed. Seems to be ok now?

PR #675, opened 2020-03-10
Don't parse ... statements

### LINKED ISSUE #674: Fix ... statements in imported modules
`...` is meant to be used as a placeholder; a stub for applications/endpoints that don't have enough information yet

given the following code: 
foo.sysl:
```
foo:
    !type fooType:
        content <: string
    fooEndpoint:
        bar <- barEndpoint
        return <: fooType
bar:
    barEndpoint:
        ...
```
bar.sysl:
```
bar:
    !type barType:
        content <: string
    barEndpoint:
        return ok <: barType
```
The Diagram shows the `...` in, even though the bar.barEndpoint have had information added, so the `...` statement should be removed from the data object. 

image

joshcarp:
Fixes #674 
If a sysl file has ... statements, it won't get passed into the sysl module (sysl.Module)
See issue  #674 for example
Given the following code

foo.sysl:
```
foo:
    !type fooType:
        content <: string
    fooEndpoint:
        bar <- barEndpoint
        return <: fooType
bar:
    barEndpoint:
        ...
```
bar.sysl:
```
bar:
    !type barType:
        content <: string
    barEndpoint:
        return ok <: barType
```

This (without change):
image
Now becomes this (with change):

image

Note: doesn't pass CI as I only want feedback on this approach before changing all the golden files


joshcarp:
Another approach is to only stop the diagram generation of the `...`, suggestions welcome

PR #676, opened 2020-03-12
#672 Make it can work well without any @package and ok<: in codegen sample model

### LINKED ISSUE #672: User simplification
Certain things with sysl are pretty tedious; 

- The need for @json_tag on every type
- The need to specify `@package` for code generation
- needing `ok <: type` in the return types 

We need either to simplify the syntax, or to have proper error messages/suggestions in the cli tool.

ericzhang6222:
Fixes #672  .

- Add code to fix issue that it can't work without @package setting in https://github.com/anz-bank/sysltemplate/blob/master/simple.sysl
- Add code to print out a warning message if model sysl is not ok <: jsonplaceholder.todosResponse, but jsonplaceholder.todosResponse

Now it can work with following new simple.sysl too:
`
import jsonplaceholder.json as jsonplaceholder ~swagger

simple:
   /foobar:
      GET:
         jsonplaceholder <- GET /todos/{id}
         return ok <: jsonplaceholder.todosResponse
`

Checklist:
- [x] Added related tests

@anz-bank/sysl-developers


joshcarp:
Please attach an example of a repo so we can see with https://github.com/anz-bank/sysltemplate

ericzhang6222:
> Please attach an example of a repo so we can see with https://github.com/anz-bank/sysltemplate

Can I update https://github.com/anz-bank/sysltemplate/blob/master/simple.sysl directly?

ericzhang6222:
> Please attach an example of a repo so we can see with https://github.com/anz-bank/sysltemplate



> > Please attach an example of a repo so we can see with https://github.com/anz-bank/sysltemplate
> 
> Can I update https://github.com/anz-bank/sysltemplate/blob/master/simple.sysl directly?

Fixed sample with PR https://github.com/anz-bank/sysltemplate/pulls

ericzhang6222:
> The change for adding a warning looks ok, but addEssentialAttr should rather be done in the transform.

@springwiz, the @package issue is when removes @pakcage information in https://github.com/anz-bank/sysltemplate/blob/master/simple.sysl, it can't generates servicehandler.go and serviceinterface.go. And I checked related transform sysl files svc_handler.sysl and svc_interface.sysl. both of them are done this package change when package information is missed in sysl model file in view goFile. So it has to change Go code still to make sure view calculation logic work well which depends on package information.

PR #677, opened 2020-03-12
Remove ... printing from sequence diagrams #major

joshcarp:
`...` statements are not printed in sequence diagrams (they are considered placeholders without any information)

Fixes #675 #674 
Changes this:
image

To This:
76368509-9d0e1480-6384-11ea-86bd-b90c79eec74e

Checklist:
- [x] Added related tests

@anz-bank/sysl-developers


nofun97:
Is it possible to have a log telling a user that there's an action that replaces `...`?

joshcarp:
What does this mean?

nofun97:
Remember how you can have the same action defined in two places but one of them is the actual action and the other is a `...`? And the one with the actual action replaces the `...`? You should warn the user that the replacement operation is happening and maybe tell them to delete the `...`.

joshcarp:
I don't want to do that; the use case of `...` is if you're stubbing out a call in a package, then importing another package. 

`...` is still needed to make individual sysl files valid by stubbing out dependencies

PR #686, opened 2020-03-16
Enum release #minor

joshcarp:
Releases enums


camh-anz:
It's a bit hard to know what you're doing here with a one word (tag) commit message. I assume #major triggers the release machinery to bump the major version? If so, why do a major release for this? It is backwards compatible. Should be a minor release.

joshcarp:
Good point

Issue #474, opened 2019-12-13
Data Diagrams don't render multiple applications in a project


joshcarp:
Given the code: 

```
Project:
    _:
        Server
        App

Server:
    !type id:
        id <: int
App:
    !type User:
        id <: Server.id
        
```
And running the command `sysl data -o project.png -j Project project.sysl`
Both `server` and `app` should be rendered, although only the last declared in `Project` is
![project](https://user-images.githubusercontent.com/32605850/70770127-45313300-1dc0-11ea-947e-10e2fb5267f1.png)
This Diagram should have both `Server.id` and `App.User`

nofun97:
https://github.com/anz-bank/sysl/blob/master/cmd/sysl/cmd_datamodel.go#L14-L71 this part seems strange because it would iterate through the applications and generate a raw plantuml for each application and then it would store the raw plantuml at `outmap[outDir]` so it replaces the processed value everytime it finishes processing

springwiz:
The correct usage of the tool as far as I remember was:
sysl data -o  "%(epname).png" -j Project data.sysl
"%(epname).png" means include all the applications in the output.

nofun97:
> The correct usage of the tool as far as I remember was:
> sysl data -o "%(epname).png" -j Project data.sysl
> "%(epname).png" means include all the applications in the output.

tried `sysl data -j Project data.sysl` with the sysl module above, since output defaults to `%(epname).png`. It still produces the same diagram

nofun97:
I discussed with @springwiz and relationships between applications are not to be represented in a data diagram, data diagram is supposed to describe relationship between tables, so by current design what @Joshcarp is doing isn't a proper use

ericzhang6222:
hi @Joshcarp, can you provide an expected diagram like https://github.com/anz-bank/sysl/issues/473? It will be very helpful, thanks.

Joshcarp:
It should be something like this: http://www.plantuml.com/plantuml/png/SoWkIImgAStDuKhEIImkLWWjJYrIu5A0qBgwTWgUALmGAI2h6YYHcQAGcLV2LSlba9gN0jG70000

ericzhang6222:
@Joshcarp, is following picture expected result?
![474_Expected](https://user-images.githubusercontent.com/59079351/72336223-4cb76500-36fb-11ea-8e1b-2f3a878799e1.png)

ericzhang6222:
[...]

PR #712, opened 2020-03-19
Allow datamodel diagrams to reference other apps

joshcarp:
Almost completes #474 
## This is a hack PR; it fixes the problem but you'll see a whole bunch of TODOS everywhere

- Allow for datamodel diagrams to be generated across applications
- NOTE: all type names now are prefixed with the application they belong to

Given the sysl file in #474 
```
Project:
    _:
        Server
        App

Server:
    !type id:
        id <: int
App:
    !type User:
        id <: Server.id
```

This now generates a file like this:
multiple-app-datamodel

Instead of the current version:
 
image

Checklist:
- [x] Added related tests
- [] Made corresponding changes to the documentation

@anz-bank/sysl-developers


pkg/datamodeldiagram/datamodel.go
```diff
@@ -86,7 +90,7 @@ func GenerateDataModelsWithPureModule(datagenParams *cmdutils.CmdContextParamDat
 }
 
 func GenerateDataModel(pclass cmdutils.ClassLabeler, outmap map[string]string, mod *sysl.Module,
-	stmts []*sysl.Statement, title, project, outDir string) {
+	stmts []*sysl.Statement, title, project, outDir string, epName bool) {
```
cuminandpaprika: Tiny bit confused here, because epName is a bool, should it be renamed to epNameMode or epNameFlag?

pkg/datamodeldiagram/datamodel.go
```diff
@@ -100,6 +104,7 @@ func GenerateDataModel(pclass cmdutils.ClassLabeler, outmap map[string]string, m
 					App:     app,
 					Title:   title,
 					Project: project,
+					Epname:  epName,
```
cuminandpaprika: Same here for epNameMode

joshcarp:
Current version of sysl generates the following
image

PR #716, opened 2020-03-20
comment statement fix

nofun97:
- Fixes `http_method_comment`

Previously `http_method_comment` requires a newline at the end which breaks the indentation rule. This PR removes that.

Before: Passes:176 Fails:57
After   : Passes:182 Fails:51

@anz-bank/sysl-developers


pkg/parser/sysl.wbnf
```diff
@@ -97,7 +97,7 @@ NativeDataTypes -> "int32" | "int64" | "int" |
 reference ->  \s* pkg=(NAME:"::" ".")? APPNAME;
 
 COMMENT_NO_NL -> "#" [^\n]*;
-COMMENT -> COMMENT_NO_NL "\n";
+COMMENT -> COMMENT_NO_NL \n+;
```
camh-anz: Why is this change here? It is not described in the PR or the commit message, which is all about `http_method_comment`. Is this change meant to be here? (it looks like it was possibly a first attempt to fix the problem).

nofun97: Right, forgot to remove this. I'll delete the change along with resolving conflict.

nofun97: change added

PR #721, opened 2020-03-23
temporary test for wbnf grammar

nofun97:
- Add temporary test for the sysl wbnf grammar. Will be replaced once a better test is implemented.

`make test-grammar` is to test which sysl files in the repository are parsed and compare it to the previous result.

`make update-grammar-result` is to update the test output once grammar change is ready.

@anz-bank/sysl-developers


Makefile
```diff
@@ -103,3 +103,11 @@ proto: pkg/sysl/sysl.pb.go
 
 help:
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+test-grammar:
+	which wbnf || go install github.com/arr-ai/wbnf/cmd
```
anz-gordonj7: this should just be `github.com/arr-ai/wbnf`

PR #723, opened 2020-03-23
Combine the ChrootFs and mod.Fs

ChloePlanet:
- [x] combine the ChrootFs and mod.Fs

@anz-bank/sysl-developers


pkg/validate/validate.go
```diff
@@ -367,16 +367,13 @@ func loadTransform(transformFile string, fs afero.Fs, p *parse.Parser) (*sysl.Ap
...
-	tokens = strings.Split(grammarFileName, ".")
-	tokens[len(tokens)-1] = "sysl"
-	grammarSyslFile := strings.Join(tokens, ".")
+	grammarFilePieces := strings.Split(tokens[len(tokens)-1], ".")
+	grammarFilePieces[len(grammarFilePieces)-1] = "sysl"
+	tokens[len(tokens)-1] = strings.Join(grammarFilePieces, ".")
 	p := parse.NewParser()
 
-	grammar, name, err := parse.LoadAndGetDefaultApp(grammarSyslFile,
-		syslutil.NewChrootFs(fs, rootGrammar), p)
```
camh-anz: I think the purpose of the `NewChrootFs` is so that any imports in the grammar sysl file are resolved relative to the grammar and also restricted to the subtree where the grammar resides. Not using a ChrootFs means the imports in the grammar may now fail because `import /foo` will import from a different path.

We will need to somehow combine the ChrootFs and ModFs. I haven't looked at ChrootFs, but i imagine that because it wraps another FS, it might work anyway, but perhaps there is a strange interaction between that and ModFs.

camh-anz: I'm not sure I agree with what I said anymore. If a `.sysl` file is using an absolute import, what should be the root for that import. If you have "import foo" in a file, it should be searched relative to the importing file, but I don't know what "import /foo" should really do. The root is usually defined outside the sysl file. Perhaps it should use the `.sysl` file logic where the presence of a file called `.sysl` defines a root. I don't know.

ChloePlanet: When explicitly add `--root`, `--root` defines a root. Otherwise `.sysl` defines the root by default. btw, `.sysl` is an empty folder according to Novan's implementation.

camh-anz:
I've merged by changes which addresses this in a slightly different way (adding a root parameter to modfs), and then making the grammar relative to the transform root. I think this PR can be closed.

pkg/mod/fs.go
```diff
@@ -24,19 +24,23 @@ func (fs *Fs) Open(name string) (afero.File, error) {
...
+		}
+
+		return syslutil.NewChrootFs(afero.NewOsFs(), mod.Dir).Open(relpath)
 	}
 
-	return syslutil.NewChrootFs(afero.NewOsFs(), mod.Dir).Open(relpath)
+	return nil, err
 }
 
 func (fs *Fs) Create(name string) (afero.File, error) {
```
camh-anz, line 46: Something I noticed when I looked at this code before...

If you embed `afero.Fs` into `mod.Fs`, you can get rid of all these functions that just delegate to the underlying `afero.Fs`:

```Go
type Fs struct {
        afero.Fs
	root   string
}

func NewFs(fs afero.Fs, root string) *Fs {
	return &Fs{Fs: syslutil.NewChrootFs(fs, root), root: root}
}
```

Then all the methods of `afero.Fs` will be promoted into `mod.Fs` and you don't need to implement the delegation manually.

In `Open()`, you would use `fs.Fs.Open(...)` instead of `fs.source.Open(...)`.

This would also increase test coverage because none of the delegations are tested.

ChloePlanet: Thanks cam! This (`fs.Fs`) is what I was looking for.

camh-anz: From https://golang.org/ref/spec#Struct_types:

> The unqualified type name acts as the field name.

Issue #684, opened 2020-03-16
Hyperlinks in svg diagrams


joshcarp:
Having the ability to link to repos/source sysl/anything would be nice within a plantuml diagram. 
An example of this is, taking this example: 

[Simple sysl data model diagram](https://anz-bank.github.io/sysl-playground/?input=UHJvamVjdDoKICAgIFNlcnZlcjoKICAgICAgICBTZXJ2ZXIKICAgIEFwcDoKICAgICAgICBBcHAgCiAgICBSZWxhdGlvbmFsLU1vZGVsOgogICAgICAgIFJlbGF0aW9uc2hpcHMKU2VydmVyOgogICAgIXR5cGUgaWQ6CiAgICAgICAgaWQgPDogaW50CkFwcDoKICAgICF0eXBlIFVzZXI6CiAgICAgICAgaWQgPDogU2VydmVyLmlkCkFub3RoZXJPbmU6CiAgICAhdHlwZSBuYW1lOgogICAgICAgIGZpcnN0TmFtZSA8OiBzdHJpbmcKICAgICAgICBsYXN0TmFtZSA8OiBzdHJpbmcKUmVsYXRpb25zaGlwczoKICAgICF0YWJsZSBBY2NvdW50OgogICAgICAgIFJlc291cmNlSUQgPDogc3RyaW5nIFt+cGtdCiAgICAgICAgQWNjb3VudElEIDw6IHN0cmluZwogICAgIXRhYmxlIFJvbGU6CiAgICAgICAgUm9sZUlEIDw6IGludCBbfnBrXQogICAgICAgIFJvbGVOYW1lIDw6IHN0cmluZwo=&cmd=c3lzbCBkYXRhIC1vICIlKGVwbmFtZSkuc3ZnIiAtaiBQcm9qZWN0IDFfcHJvamVjdC5zeXNs)

[...]

PR #728, opened 2020-03-24
Hyperlinks on appname and endpoint for sequence diagrams

ashwinsajiv:
Addresses issue #684 
- Adds hyperlinks to appnames and endpoints which has the link attribute mentioned.


demo/simple/sysl-app-hyperlink.sysl
```diff
@@ -0,0 +1,12 @@
+MobileApp:
+    @link = "www.foofle.com"
```
joshcarp: Might want to add that this needs to have https:// trailing otherwise browsers think it's a relative link (just tried it now)

ashwinsajiv: Addressed

ashwinsajiv: Addressed

PR #729, opened 2020-03-24
Wbnf grammar fixes

nofun97:
Multiple small fixes on the grammar. I have divided them into multiple commits for easier review. I can divide them into multiple PRs but the fixes are very small.

- `union` rule has working `COMMENT` statement.
- `union` rule for `user_defined_type` is changed to `NAME`. `type_spec` allows sized types which isn't allowed in the original grammar.
- `param` rule now allows newline inside parameters. Current `sysl validate` doesn't allow this, but this needs to be allowed as one of the allowed argument is an field with multiple attributes. For example:
```
field <:
      attribute1 <: int
      attribute2 <: string
```
- `field` multiple attributes now work like the example above.
- `alias` restricts to one `type` line (it allowed multiple `type` line before)
- `annotation` now allows full stop.

@anz-bank/sysl-developers


nofun97:
added a test case for fields with multiple attributes

PR #735, opened 2020-03-25
handles lines with random spaces between indented statements

nofun97:
current indentation rule does not anticipate lines with random spaces. For example (with underscores replacing spaces)

```
____statement1
___

____statement2
```

The newly added regex rule `/{(\n (?:[\_\t]+ \n)?)+}` will handle lines with random spaces.
It is also added to the global `.wrapRE` rule so that it does not ignore the random spaces as sometimes it is parsed as the `level` rule which consumes spaces before a statement is necessary to determine the level of indentation of the statement.

Parse result is also updated along with the newly added sysl files.

@anz-bank/sysl-developers

### DIFF
--- a/scripts/grammar-out.txt
+++ b/scripts/grammar-out.txt
@@ -4,16 +4,19 @@
 ./demo/simple/reljam-xsd.sysl success
 ./demo/simple/sysl-sd.sysl success
 ./demo/simple/example2.sysl success
+./demo/simple/sysl-app-hyperlink.sysl success
 ./demo/simple/hello.sysl success
+./demo/simple/sysl-endpoint-hyperlink.sysl success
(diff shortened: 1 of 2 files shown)


scripts/grammar-out.txt
```diff
@@ -4,16 +4,19 @@
 ./demo/simple/reljam-xsd.sysl success
 ./demo/simple/sysl-sd.sysl success
 ./demo/simple/example2.sysl success
+./demo/simple/sysl-app-hyperlink.sysl success
```
anzdaddy: As a separate PR, can we move outcome to the front and first sort by that?

nofun97: I can try, I'll make a separate PR

PR #736, opened 2020-03-26
Add IntelliJ .idea folder to gitignore

andrewemeryanz:


andrewemeryanz:
> And yes, I get that this means everyone has to do it for themselves, but the alternative is having ignores for multiple IDEs, editors, etc., in every repo.

Ok. I'm fine with this approach.

PR #737, opened 2020-03-26
[WIP] Introduce --parser command-line argument to enable wbnf parsing

andrewemeryanz:
Proposal:
Introduce --parser command-line argument to enable wbnf parsing

To simplify the testing and transition to wbnf, the --parser flag can
switch between the existing Antlr parser and the under-development
wbnf parser.

Usage:
sysl validate example.sysl --parser wbnf

@anz-bank/sysl-developers

### DIFF
--- a/cmd/sysl/cmd_codegen.go
+++ b/cmd/sysl/cmd_codegen.go
@@ -54,6 +54,7 @@ func (p *codegenCmd) Execute(args cmdutils.ExecuteArgs) error {
 			Start:         p.Start,
 			DepPath:       p.DepPath,
 			BasePath:      p.BasePath,
+			ParserType:    args.ParserType,
 			Filesystem:    args.Filesystem,
 			Logger:        args.Logger,
 		})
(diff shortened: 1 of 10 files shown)


andrewemeryanz:
> Right now there is no wbnf parser, so adding the scaffolding to switch the runtime parser seems to be jumping the gun a bit, and then, when there is a wbnf parser we would almost certainly want to simply switch the parser out without the users knowing anything changed, so we would want to then undo this sort of change.

With a perfect transition, yes. I'm foreseeing a desire to start consuming the wbnf parser before it's at feature parity. It may never get to feature parity. My assumption is also that there will be a desire to invoke the Antlr parser once the switch has been made to compare against unexpected quirks in existing behaviour.

> The PR That I raised on the other hand allows us to build out the parser and get live feedback (via the textpb diff) comparing the antrl and wbnf parser outputs. I think concentrating on the wbnf parser implementation seperate from the existing sysl codebase is a much safer direction.

Comparing parser outputs is a good approach. This change doesn't preclude this. This change provides an integration point for early integration. Other than exposing a command-line argument that may be deprecated in the future, I see nothing unsafe about this approach.

andrewemeryanz:
Stale.

Issue #722, opened 2020-03-23
Invalid dependency: protoc-gen-validate


andrewemeryanz:
### Description
Invalid dependency `protoc-gen-validate` when resolving dependencies.

### Steps to Reproduce
1. Run:
```shell
go test ./...
```

**Expected behavior:**
Tests run.

**Actual behavior:**
```shell
go: github.com/envoyproxy/protoc-gen-validate@v0.3.0: reading github.com/envoyproxy/protoc-gen-validate/go.mod at revision v0.3.0: unknown revision v0.3.0
```

### Environment
Sysl version: 0.17.0

andrewemeryanz:
Related: https://github.com/envoyproxy/protoc-gen-validate/issues/325

PR #742, opened 2020-03-27
Removes invalid protoc-gen-validate dependency

andrewemeryanz:
No tagged version of protoc-gen-validate:0.3.0 exists. The closest valid
version: protoc-gen-validate:0.3.0-java appears unrequired:

go mod why github.com/envoyproxy/protoc-gen-validate
main module does not need package github.com/envoyproxy/protoc-gen-validate

Closes: #722

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,6 @@ require (
 	github.com/cornelk/hashmap v1.0.1
 	github.com/dchest/siphash v1.2.1 // indirect
 	github.com/dlclark/regexp2 v1.2.0 // indirect
-	github.com/envoyproxy/protoc-gen-validate v0.3.0 // indirect
 	github.com/fullstorydev/grpcui v0.2.1
 	github.com/fullstorydev/grpcurl v1.5.0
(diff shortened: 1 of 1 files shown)


joshcarp:
Oh yeah, I had a quick thought on this before; I've forked the repo to https://github.com/anz-bank/protoc-gen-validate so we won't need to worry about tags getting deleted; what do you think about this?

joshcarp:
was this done via `go mod tidy`; I'm just figuring out if this is going to break some of the work i've been doing because that's where I first got this bug even though it didn't appear in the main sysl branch

andrewemeryanz:
> was this done via go mod tidy

No. If go can't resolve the module then it fails before attempting to tidy.

> I've forked the repo ... what do you think about this?

As a general approach I'd like to think this wasn't necessary, but as a temporary measure, sure. It was an indirect dependency. I had a look through the codebases of other dependencies that might have used this themselves but I couldn't find where it was used... so no definitive answer as to how it got there.

PR #749, opened 2020-03-30
Fix openapi import when colon is in URL path

cuminandpaprika:
- Adds escaping of ":" and "+" in URL paths when importing from OpenAPI or Swagger.

e.g
`/api:pay` in the OpenAPI yaml will be imported as '/api%3A' in sysl

Fixes https://github.com/anzx/acceleration/issues/58

### DIFF
--- a/pkg/importer/utils.go
+++ b/pkg/importer/utils.go
@@ -72,7 +72,7 @@ func spaceSeparate(items ...string) string {
 }
 
 func getSyslSafeEndpoint(endpoint string) string {
-	// url.PathEscape does not escape '.'
+	// url.PathEscape does not escape '. :'
 	charsToKeep := map[string]string{
 		`%2F`: "/",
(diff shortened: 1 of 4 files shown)


pkg/importer/utils.go
```diff
@@ -83,6 +83,8 @@ func getSyslSafeEndpoint(endpoint string) string {
 	}
 	endpoint = url.PathEscape(endpoint)
 	endpoint = strings.ReplaceAll(endpoint, ".", `%2E`)
+	endpoint = strings.ReplaceAll(endpoint, ":", `%3A`)
+	endpoint = strings.ReplaceAll(endpoint, "+", `%2B`)
```
nofun97, line 87: should collect the characters in a map, like `charsToEscape`

cuminandpaprika: nice, fixed up!

PR #750, opened 2020-03-30
Change modules to make the more usable for codegen

camh-anz:
Add a `root` parameter to `mod.Fs` so that when a root is specified, it can
be a module.

When the `codegen` subcommand loads a grammar (from the `--grammar` flag),
look for it relative to `--root-transform` instead of `--root`.

The grammar is closely tied to the transform so the transform root is the
more correct place to look for it.

This allows the following invocation:

    sysl codegen --root-transform github.com/anz-bank/sysl-go/codegen \
    --grammar grammar/go.gen.g --transform transform/svc_types.sysl \
    ...

@anz-bank/sysl-developers

### DIFF
--- a/pkg/mod/fs.go
+++ b/pkg/mod/fs.go
@@ -11,10 +11,11 @@ import (
 
 type Fs struct {
 	source afero.Fs
+	root   string
 }
 
-func NewFs(fs afero.Fs) *Fs {
-	return &Fs{source: fs}
+func NewFs(fs afero.Fs, root string) *Fs {
+	return &Fs{source: fs, root: root}
 }
 
 func (fs *Fs) Open(name string) (afero.File, error) {
@@ -23,6 +24,9 @@ func (fs *Fs) Open(name string) (afero.File, error) {
 		return f, nil
 	}
 
+	// filepath.Join will strip path elements of ".", so if the root is "."
+	// it will still work as a go module path when prepended with "."
+	name = filepath.Join(fs.root, name)
 	mod, err := Find(name)
 	if err != nil {
 		return nil, err
@@ -53,6 +57,9 @@ func (fs *Fs) OpenFile(name string, flag int, perm os.FileMode) (afero.File, err
 		return f, nil
 	}
 
+	// filepath.Join will strip path elements of ".", so if the root is "."
+	// it will still work as a go module path when prepended with "."
+	name = filepath.Join(fs.root, name)
 	mod, err := Find(name)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


pkg/mod/fs.go
```diff
@@ -23,6 +24,7 @@ func (fs *Fs) Open(name string) (afero.File, error) {
 		return f, nil
 	}
 
+	name = fs.root + string(os.PathSeparator) + name
```
anz-gordonj7: filepath.Join() here and below

camh-anz: I was going to say for that I need to create a slice, so I don't want to do that, but that's `strings.Join()`.

Will do.

camh-anz: Fixed in 40dc688

pkg/validate/validate.go
```diff
@@ -367,18 +367,21 @@ func loadTransform(transformFile string, fs afero.Fs, p *parse.Parser) (*sysl.Ap
...
 	}
 	return grammar.GetApps()[name], nil
 }
+
+func changeFileExtension(filename, extension string) string {
+	parts := strings.Split(filename, ".")
+	if len(parts) == 1 {
+		parts = append(parts, "") // add empty extension to be replaced.
+	}
+	parts[len(parts)-1] = "sysl"
```
ChloePlanet:
```suggestion
	parts[len(parts)-1] = extension
```

camh-anz: Good catch.

Fixed in 79c500b62912afc1ba40ad4c7f1c69c1f4ed7127

gkanz:
The changes won't break anything and do align with the current relationship between transforms and grammars.

I have the following reservation but I don't think it is sufficient to reject the PR:
It is conceivable that there will be in the future multiple sets of transforms (from different modules) that all rely on the same grammar. While transforms are tightly coupled to grammars, grammars are not tightly couple to transforms.

camh-anz:
> It is conceivable that there will be in the future multiple sets of transforms (from different modules) that all rely on the same grammar. While transforms are tightly coupled to grammars, grammars are not tightly couple to transforms.

I suspect the new arrai-based templating codegen may well be ready by that future point. Even if not, I imagine we could add a `--grammar-root` at that point if a third root makes sense

camh-anz:
Hmmm, this only works because `filepath.Join()` cleans the path, so `./foo` becomes `foo`. That means with the default `--root-transform` of `.`, when that is prefixed to the transform/grammar filename when using modules, `./github.com/anz-bank/sysl-go` becomes `github.com/anz-bank/sysl-go`. This is what we want but it is mostly because of a side-effect. Not ideal.

One option is to make the default for `--root-transform` to be `` (empty string) instead of `.`. The `ChrootFs` handles a root of empty as `.`, but perhaps that is also a bit of a side-effect.

Prior to modules, all paths were considered filesystem paths, so a default of `.` make sense. But now we have modules, a path may be an external reference, so it does not make sense to prefix `.` in those cases.

camh-anz:
> This is what we want but it is mostly because of a side-effect. Not ideal.

Added a comment so it is not so non-obvious.

PR #756, opened 2020-04-02
Sequence diagram generation using mermaidjs

### LINKED ISSUE #433: Consider using https://mermaidjs.github.io/#/
Possible alternative to plantuml.

ashwinsajiv:
- Mermaidjs generated for the given sysl
- Only deals with calls and returns in the sysl file
- Added test cases for simple sysl to mermaid generation

Partially Fixes #433 

@anz-bank/sysl-developers


pkg/mermaidsequencediagram/mermaid.go
```diff
@@ -0,0 +1,79 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/anz-bank/sysl/pkg/sysl"
+)
+
+const projDir = "../../"
+
+func GenerateMermaidSequenceDiagram(m *sysl.Module, appname string, epname string,
+	previousapp string, indent int, thestart bool) (string, error) {
```
joshcarp, line 13: `thestart` is a bool but every test case you have sets it to `true` - do you think it's worth getting rid of this parameter to simplify the API?

ashwinsajiv: thestart is set to true only once from the beginning of the function. It is very much needed cos I am recursively calling the same function and I would want to know when I've reach the start again for implementing the call/return to "the world".

ashwinsajiv:
> This is awesome!
> 
> It would be nice to see some comments on functions; I can see this package be used outside of sysl

Will add comments once I have completed the whole thing

pkg/mermaidsequencediagram/mermaid.go
ashwinsajiv: Check line 32

PR #760, opened 2020-04-02
Consolidate workflows into one

cuminandpaprika:
This merges the github action workflow from each platform into a single workflow so that common actions do not need to be duplicated.
This means that changes made to CI can be done in a single place, for all target platforms

- Removes codecov
- Consolidates workflow files into one, to reduce duplication

Checklist:
- [x] Added related tests

@anz-bank/sysl-developers


.github/workflows/go.yml
```diff
@@ -0,0 +1,155 @@
...
+      - name: Set GOVERSION environment variable (Linux/macOS)
+        if: matrix.os != 'windows-latest'
+        run: echo ::set-env name=GOVERSION::$(go version | awk '{print $3, $4;}')
+
+      - name: Set GOVERSION environment variable (Windows)
+        if: matrix.os == 'windows-latest'
+        run: $Env:GOVERSION=(go version | %{($_ -split " ")[-2,-1] -join " "}); echo $Env:GOVERSION
+
+      - name: Build sysl binary via goreleaser (Linux)
+        if: matrix.os != 'windows-latest'
```
camh-anz, line 66: should be `matrix.os == 'ubuntu-latest'`

.github/workflows/go.yml
```diff
@@ -0,0 +1,155 @@
...
+      - name: Test (Linux/macOS)
+        if: matrix.os != 'windows-latest'
+        run: make coverage
+        timeout-minutes: 3
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SYSL_PLANTUML: http://www.plantuml.com/plantuml
+
+      - name: Test (Windows)
+        if: matrix.os == 'windows-latest'
```
camh-anz, line 50: Making these all steps instead of jobs means they will all run in series. This is going to slow down the build and make it take longer to merge PRs. I think you should at least keep windows as a separate job as it shares no steps with the unix-like build.

I'm not sure how the matrix interacts with steps, but if every step has a !windows or windows gate, there's little point putting it into a matrix.

camh-anz: ok, so I read up on the github build matrix, and each runs as a separate job, so I guess my objection is moot, but this still reads awkwardly with all the intermixed OS steps. Kind of makes a mockery of the matrix stuff. I wonder if there's a better way - could we include the common steps into other jobs somehow, or parameterise the os stuff in a makefile or something? thinking out aloud.

cuminandpaprika: Yup, in this case, they do run in parallel. I agree with the readability part though, it suffers cause of the branching. I’ll take a look and see if there’s ant other way but I think the reduction in duplication here is worth it

camh-anz: The duplication can be reduced to:
```
      - name: Set up Go
        uses: actions/setup-go@v1
        with:
          go-version: 1.13
      - name: Check out code into the Go module directory
        uses: actions/checkout@master
        # Needed for changelog generation by goreleaser
      - name: Unshallow
        run: git fetch --prune --unshallow
```

This also implies that the `go mod tidy` (`make deps`) check is done in a different job (without the unshallow as it is not needed). I think a different job makes sense for that as it is a separate gate. Also the `make deps` part is not really needed. Each step will download the deps it needs as it builds, so you don't need the `go get ...` stuff explicitly.

camh-anz: Also, I think https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches is saying you can do the unshallow in the same step as the checkout.

.github/workflows/go.yml
```diff
@@ -0,0 +1,155 @@
...
+        os: [ubuntu-latest, macOS-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+        id: go
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
cuminandpaprika: Aha! Thanks for that, will do!

.github/workflows/go.yml
cuminandpaprika: Had a look at the options, and it looks like there's 2 options for making it more readable:
1. Put all the conditional steps into a makefile (would make the workflow more readable, as the workflow would just do the sequencing, and the makefile provides the actual task definition)
2. Put the common steps in their own workflow, and maintain the 3 separate workflow files (This seems really messy, since you'd end up with multiple workflow files, might not work since subsequent steps are dependant on previous ones. You'd have to share the files/ artifacts in some way ) 

I'm learning towards option 1 more, but I'm hoping that I can do that in a seperate PR, along with a more comprehensive cleanup of the Makefile. Thoughts?

camh-anz: do they need to be separate workflow files? - there is already two jobs in the one file here. can't you do the same with the build/test jobs?

.github/workflows/go.yml
```diff
@@ -0,0 +1,155 @@
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macOS-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+        id: go
```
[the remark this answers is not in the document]
cuminandpaprika: Double checked and this doesn't seem to get used anywhere, I've removed it

.github/workflows/go.yml
cuminandpaprika: Yes indeed, fixed!

.github/workflows/go.yml
cuminandpaprika: hmm, I think you can run them as dependant jobs, so they don't run in parallel, but you'd still have similar difficulties with sharing files etc, as each job runs in its own environment.

Definitely seeing the limitations of trying to build a workflow here with a YAML based DSL here!

camh-anz: i dont think there is any sharing of files across jobs whether you use a matrix or not.

.github/workflows/go.yml
```diff
@@ -0,0 +1,131 @@
+name: Build and Test
+
+on:
+  push:
+    branches:
```
camh-anz, line 5: Since this is "all branches" I think we can just leave this out. Also on the `pull_request` section.

.github/workflows/go.yml
```diff
@@ -0,0 +1,131 @@
...
+        with:
+          go-version: 1.13
+        
+        # --prune --unshallow is needed for changelog generation by goreleaser
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        run: |
+          git fetch --prune --unshallow
+        
+      - name: Unshallow
```
camh-anz, line 33: this step is not needed any more.

.github/workflows/go.yml
```diff
@@ -0,0 +1,131 @@
...
+      matrix:
+        os: [ubuntu-latest, macOS-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+        
+        # --prune --unshallow is needed for changelog generation by goreleaser
```
camh-anz, line 27: I would put this comment on the line that does the fetch: `# needed for goreleased changelog generation`

.github/workflows/go.yml
cuminandpaprika: Thanks, fixed!

.github/workflows/go.yml
cuminandpaprika: 👍

.github/workflows/go.yml
cuminandpaprika: yup, done

.github/workflows/go.yml
cuminandpaprika: Okay, so it turns out the examples from https://github.com/actions/checkout#Fetch-all-history-for-all-tags-and-branches seem to be wrong or outdated. I got the following message when trying to do use 

```
      - name: Check out code into the Go module directory
        uses: actions/checkout@v2
        # needed for changelog generation by goreleaser
        run: |
          git fetch --prune --unshallow 
```

> a step cannot have both the uses and run keys

camh-anz:
It looks like you will not be able to merge this because the required status checks are on the old build steps.

PR #761, opened 2020-04-02
Add go mod tidy check

cuminandpaprika:
Adds a check to ensure go.mod and go.sum files are tidy

Tested here: https://github.com/anz-bank/sysl/pull/773/checks?check_run_id=566134662

Checklist:
- [x] Added related tests

@anz-bank/sysl-developers

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -25,9 +25,6 @@ jobs:
       - name: Unshallow
         run: git fetch --prune --unshallow
 
-      - name: Get dependencies
-        run: make deps
-
       - name: Test
         run: make coverage
         timeout-minutes: 3
@@ -80,9 +77,6 @@ jobs:
       # Needed for changelog generation by goreleaser
     - name: Unshallow
       run: git fetch --prune --unshallow
-
-    - name: Get dependencies
-      run: make deps
 
     - name: Test
       run: .\scripts\test-with-coverage-windows.bat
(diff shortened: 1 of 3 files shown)


.github/workflows/go.yml
```diff
@@ -0,0 +1,155 @@
...
+      - name: Set GOVERSION environment variable (Linux/macOS)
+        if: matrix.os != 'windows-latest'
+        run: echo ::set-env name=GOVERSION::$(go version | awk '{print $3, $4;}')
+
+      - name: Set GOVERSION environment variable (Windows)
+        if: matrix.os == 'windows-latest'
+        run: $Env:GOVERSION=(go version | %{($_ -split " ")[-2,-1] -join " "}); echo $Env:GOVERSION
+
+      - name: Build sysl binary via goreleaser (Linux)
+        if: matrix.os != 'windows-latest'
```
camh-anz, line 61: This step says "(Linux)" but the matrix test says "not windows". I think this is meant to be `== ubuntu-latest`?

.github/workflows/go.yml
```diff
@@ -0,0 +1,155 @@
...
+      - '*'
+
+jobs:
+  build-and-test:
+    name: Build Go and Test
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macOS-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
```
camh-anz, line 18: from a brief internet search, steps cannot run in parallel, only jobs. So by putting this all in one job, the windows, mac and linux builds are now sequential instead of parallel. This will slow down CI by quite a bit I'd imagine.

I think it will be faster to split these into separate jobs and also clearer as all the steps are mixing up the different OS steps.

Issue #755, opened 2020-04-01
Sysl release workflow is not working


ChloePlanet:
### Description
The sysl release machinery is not working. When building the release, `go.sum` is getting modified so the workspace is identified as dirty, which [breaks the release](https://github.com/anz-bank/sysl/runs/538539415?check_suite_focus=true). This is why we have not had a release since v0.29.0.

### Steps to Reproduce
1. merge PR to master
* Debug&Crash Output
```
   ⨯ release failed after 0.29s error=git is currently in a dirty state, please check in your pipeline what can be changing the following files:
 M go.sum
```

cuminandpaprika:
First issue:
- go sum file is dirty fixed in https://github.com/anz-bank/sysl/pull/754

Next issue:
The makefile isn't getting copied over correctly when the docker image is built using goreleaser
Step 6/11 : COPY . .
 ---> 64f086f7a424
Step 7/11 : RUN make build
 ---> Running in 747c75dbccc6
make: *** No rule to make target 'build'.  Stop.
The command '/bin/sh -c make build' returned a non-zero code: 2
: exit status 2

cuminandpaprika:
Steps
- [ ] Merge workflows into one
- [ ] Add check-tidy so that branches with dirty go.mod and go.sum can't be merged in
- [ ] Build and upload the docker image using a github action

PR #769, opened 2020-04-06
Build docker image with workflow step

cuminandpaprika:
Removes the docker image build step from goreleaser and executes it using the offical docker github action

Fixes https://github.com/anz-bank/sysl/issues/755

To view a test run, I've run it on my fork, pushing a docker image to https://hub.docker.com/repository/registry-1.docker.io/zhengj9/test-sysl/tags?page=1
Here is the execution of the github action https://github.com/cuminandpaprika/sysl/runs/563566184

The docker image can be tested using
`docker run zhengj9/test-sysl:latest info`

- Add build step in the release workflow
- Add docker build job to the build and test workflow

Checklist:
- [x] Added related tests

@anz-bank/sysl-developers

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -124,3 +124,27 @@ jobs:
           npm test --prefix unsorted/sysl_js
         env:
           NPM_AUTH_TOKEN: "SOME-RANDOM-KEY"
+
+  build-docker:
+    name: Builds the sysl docker image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go 1.13
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+
+      - name: Check out code into the Go module directory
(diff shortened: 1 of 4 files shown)


.github/workflows/go.yml
```diff
@@ -124,3 +124,27 @@ jobs:
...
+    steps:
+      - name: Set up Go 1.13
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@master
+
+      - name: Build and push docker image
```
ChloePlanet, line 140: Why push docker image in go-test workflow?

cuminandpaprika: This only builds it in the test workflow, to ensure that the docker image is still buildable.

If you look at like 149, push is set to false, so the image won't be published, only built

PR #790, opened 2020-04-15
Add .link and .html outputs

joshcarp:
adds .html outputs that output html img tags  and .links that output raw links


cmd/sysl/cmd_datamodel_test.go
```diff
@@ -49,3 +54,81 @@ func TestDataModel(t *testing.T) {
...
+}
+
+func TestLinkOutput(t *testing.T) {
+	t.Parallel()
+	p := &datamodelCmd{}
+	p.Output = "whatever.html"
+	p.Project = "Project"
+	p.ClassFormat = "%(classname)"
+	p.Direct = false
+	planutml := os.Getenv("SYSL_PLANTUML")
```
cuminandpaprika, line 104: Typo

joshcarp:
Yeah maybe, I'm only using this as an experiment to see how fast I can get diagram generation in sysl-catalog. 
If there seems to be more use for this (I can also think of the sysl-playground) I'll add it to the docs.

PR #793, opened 2020-04-17
Add ability to hide calls from sequence diagram generation

joshcarp:
Allows for hidden calls in sequence diagrams:
image

This is achived by adding ~hidden to the endpoint so it won't show up in sequence diagram generation:
```
App:
    Endpoint1:
        SecondApp <- Endpoint3
SecondApp:
    # Any calls to Endpoint3 will be hidden in diagram generation, but whatever calls it contains will still exist
    Endpoint3[~hidden]:
        return whatever

```


pkg/sequencediagram/sequencediagram_test.go
```diff
@@ -61,6 +61,45 @@ deactivate _0
...
+	l := &cmdutils.Labeler{}
+	p := &SequenceDiagParam{Title: "Profile"}
+	p.Endpoints = []string{"App <- Endpoint1"}
+	p.AppLabeler = l
+	p.EndpointLabeler = l
+	r, err := GenerateSequenceDiag(m, p, logger)
+	assert.NotNil(t, m)
+	assert.NotNil(t, r)
+	assert.Nil(t, err)
+	expected := `''''''''''''''''''''''''''''''''''''''''''
```
cuminandpaprika, line 80: Thoughts on doing a `assert.True(t, strings.Contains(r, "Endpoint3"))` instead? That means you can easily change the input sysl file in future tests

joshcarp: I'd rather have a rigid testcase that requires an exact match, otherwise there's the risk that the testcase "fails" but doesn't actually fail the test

Issue #786, opened 2020-04-14
Refine utility package and cleanup ad hoc functions


joshcarp:
There are a lot of functions to deal with sysl proto objects (like processing names/getting type declarations and stuff) which have been redeclared in many packages as engineers often don't know that it's probably been written elsewhere. 
https://github.com/anz-bank/sysl/tree/master/pkg/syslutil

Stuff like `GetAppName` and functions surrounding type handling and evaluation [here](https://github.com/anz-bank/sysl/blob/55dfdb6d7494bb34d4c20e6fa3b0c688a50788f2/pkg/syslutil/helpers.go#L21)
should all go into the syslutil package and should also be heavily documented to avoid duplicate code. 

All of this should be moved to pkg/syslutil in order to ensure standard behaviour across all of the sysl codebase, and this can also be thought of as the first step into actually simplifying the sysl proto and parsing so that utility functions don't even need to be created to start with.

cuminandpaprika:
Awesome writeup, exactly what I had in mind. Syslutil does seem like a useful place to have it, but in the long term, I think the “clean” data object that we build in Syslutil should become the return type from the Parser.

cuminandpaprika:
Okay so here's the inital type that's a cleaned up version of sysl's protobuf types

```

         "MegaDatabase": {
          "Name": "MegaDatabase",
          "Attributes": {
           "package": "ServerPackage"
          },
          "Endpoints": {},
          "Types": {
           "AnotherEmpty": {
            "Source": "",
            "Type": "",
            "Items": null,
            "Enum": null,
            "Properties": null
           },
           "Empty": {
            "Source": "",
            "Type": "",
            "Items": null,
            "Enum": null,
            "Properties": null
           },
           "Money": {
            "Source": "",
            "Type": "tuple",
            "Items": null,
            "Enum": null,
            "Properties": {
             "amount": {
              "Source": "",
              "Type": "int",
              "Items": null,
              "Enum": null,
              "Properties": null
             }
            }
           }
          }
         },
         "MobileApp": {
          "Name": "MobileApp",
          "Attributes": {
           "package": "ApplicationPackage"
          },
          "Endpoints": {
           "Login": {
            "Path": "Login",
            "Params": {
             "input": {
              "Name": "input",
              "Type": {
               "Source": "",
               "Type": "tuple",
               "Items": null,
               "Enum": null,
               "Properties": {
                "query": {
                 "Source": "",
                 "Type": "string",
                 "Items": null,
[...]

PR #799, opened 2020-04-20
Add sysl wrapper package WIP

cuminandpaprika:
This PR adds a syslwrapper package which abstracts some of the complex details of the sysl.App data type.

This is used by an openapi3 exporter PR which will be coming up.

It's not a complete implementation yet, and still needs some more work, which will be coming in future PRs.

- [x] Add wrapper type for sysl.Application
- [x] Implement type resolution

To be done in future PRs
Fix https://github.com/anz-bank/sysl/issues/786.
- [ ] Add dependency info 
- [ ] Consolidate helper functions for creating sysl applications
- [ ] Create test util for inline Sysl
- [ ] Tidy util functions

Checklist:
- [x] Added related tests

### DIFF
--- a/pkg/syslwrapper/app_test.go
+++ b/pkg/syslwrapper/app_test.go
@@ -7,74 +7,12 @@ import (
 
 	"github.com/anz-bank/sysl/pkg/parse"
 	"github.com/anz-bank/sysl/pkg/sysl"
-	"github.com/anz-bank/sysl/pkg/syslutil"
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
 )
 
-func readSyslModule(filename string) (*sysl.Module, error) {
-	mod, err := parse.NewParser().Parse(filename,
-		syslutil.NewChrootFs(afero.NewOsFs(), "."))
-	return mod, err
-}
-
-func prettyPrint(t *testing.T, v interface{}) {
-	json, err := json.MarshalIndent(v, "", " ")
-	if err != nil {
-		t.Log(t, err)
-	}
-	t.Log(t, json)
-}
-
-// GenerateTypeGraph takes in a simplified sysl type and resolves the sub type references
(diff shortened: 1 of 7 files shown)


anz-gordonj7:
I struggle to see the benefits from the api you've defined over using the raw protobuf types.

What problem are you actually trying to solve with this work? I would simply clean up the helper functions that exist, add more as needed and potentially look into an `interface` style API instead of a `struct` one, this proposal will require a massive conversion effort over the entire codebase in one hit, with a more simplified refactoring approach both old and "new" style API could co-exist.

cuminandpaprika:
> I struggle to see the benefits from the api you've defined over using the raw protobuf types.
> 
> What problem are you actually trying to solve with this work? I would simply clean up the helper functions that exist, add more as needed and potentially look into an `interface` style API instead of a `struct` one, this proposal will require a massive conversion effort over the entire codebase in one hit, with a more simplified refactoring approach both old and "new" style API could co-exist.

Thanks for the feedback @anz-gordonj7, the problem this is trying to solve is the fact that many of the packages (such as the diagram generation packages) in the Sysl library need large amounts of data wrangling when using the raw protbuf types. I'm hoping to abstract away some of the complexity inherent in the raw protobuf type.

I definitely agree an interface style API could serve better here, so that both styles can be supported. Will try that approach here

pkg/syslwrapper/app_test.go
```diff
@@ -0,0 +1,448 @@
...
+	"testing"
+
+	"github.com/anz-bank/sysl/pkg/parse"
+	"github.com/anz-bank/sysl/pkg/sysl"
+	"github.com/anz-bank/sysl/pkg/syslutil"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+)
+
+func readSyslModule(filename string) (*sysl.Module, error) {
```
joshcarp, line 15: this function is the same as ` parse.NewParser().Parse(filename, afero.NewOsFs())`

cuminandpaprika: Cheers, will replace that!

PR #800, opened 2020-04-20
RFC: Add a simplified sysl command to orchastrate codegen from a sysl file

anz-gordonj7:
If you have a model.sysl which already defines your endpoints and you
just want to use the sysl-go rest transforms adding something like:
```
Codegen [~codegen]:
    @grammar="grammars/go.gen.g"
    @grammar_start="goFile"
    rest_frontend:
        @targets=["transforms/svc_types.sysl", "transforms/svc_router.sysl", "transforms/svc_handler.sysl", "transforms/svc_interface.sysl"]
    rest_backend:
        @grammar="grammar/go.gen.g"
        @targets=["transforms/svc_types.sysl", "transforms/svc_error_types.sysl", "transforms/svc_client.sysl"]

```

to the model will automatically run all the relevant transforms for each app with special annotations defined.

```

Bff:
    @codegen_transform="rest_frontend"
    @codegen_dirname="bff"
    @codegen_basepath="github.service.anz/dcx/anzapp-bff/codegen/gen"

```

The above will tell the new codegen command to run all the `rest_frontend` targets over the `Bff` app and
write the result to the `$outdir/bff` directory.

Unscientific tests with the dcx/bff sysl project dropped codegen runtime from 1min 40s to 1min 24sec (consistant)
There is also definitly room for optimisation by not reloading the transform .sysl (and grammar) multiple times.

- new command `sysl syslgen --outdir=$dir <model>`

------

This is more of a proof of concept than a final implementation. I'm not 100% convinced with the required sysl annotations (and using simple endpoints as the target names), would love some discussion

See https://github.service.anz/gist/gordonj7/6a0a80cdebde6d183b9ebeb6207c2c0b for a more full example

### DIFF
--- a/cmd/sysl/cmd_codegen2.go
+++ b/cmd/sysl/cmd_codegen2.go
@@ -1,0 +1,132 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+
+	"github.com/anz-bank/sysl/pkg/sysl"
+
+	"github.com/anz-bank/sysl/pkg/cmdutils"
+
+	"github.com/anz-bank/sysl/pkg/syslutil"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+type codegen2Cmd struct {
+	cmdutils.CmdContextParamCodegen
+	outDir         string
+	appName        string
+	validateOnly   bool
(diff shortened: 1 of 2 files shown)


joshcarp:
Why does this need to be in a different command?

anz-gordonj7:
> Why does this need to be in a different command?

it doesnt really, I just didnt want to chnge the existing command.

PR #822, opened 2020-04-29
Add printer package from joshcarp/sysl-printer

joshcarp:
Adds a package that can print out various sysl data objects back to source code. 

see https://github.com/joshcarp/sysl/blob/sysl-printer2/pkg/printer/README.md for more details

Used in protoc-gen-sysl


pkg/printer/printer.go
```diff
@@ -0,0 +1,251 @@
...
+type Printer struct {
+	io.Writer
+}
+
+// NewPrinter returns a printer that can be used to print out sysl source code from data structures
+func NewPrinter(buf io.Writer) *Printer {
+	return &Printer{Writer: buf}
+}
+
+// PrintModule Prints a whole module, calling
```
cuminandpaprika, line 30:
```suggestion
// PrintModule Prints a whole module
```

pkg/printer/alphabetical.go
```diff
@@ -0,0 +1,58 @@
+// alphabetical.go are the functions to get a sorted list of keys from sysls map types
```
cuminandpaprika, line 1:
```suggestion
// alphabetical.go contains functions which return a sorted list of keys from sysl map types
```

pkg/printer/README.md
```diff
@@ -0,0 +1,42 @@
+# printer
```
cuminandpaprika, line 1: Love having READMEs and usage instructions for packages! We should adopt this for all new packages written for sysl!

pkg/printer/printer_test.go
```diff
@@ -0,0 +1,56 @@
...
+	log := logrus.Logger{}
+
+	module, _, err := loader.LoadSyslModule("/", "printer.sysl", fs, &log)
+	assert.NoError(t, err)
+
+	fileBytes, err := afero.ReadFile(fs, "printer.sysl")
+	assert.NoError(t, err)
+	var buf bytes.Buffer
+	pr := NewPrinter(&buf)
+	//pr := NewPrinter(os.Stdout)
```
cuminandpaprika, line 29: Could probably remove this line?

pkg/printer/alphabetical.go
```diff
@@ -0,0 +1,58 @@
...
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return keys
+}
+
+func alphabeticalEndpoints(m map[string]*sysl.Endpoint) []string {
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		if k != "" {
```
cuminandpaprika, line 34: Any reason why the empty string check only needs to be done for endpoint names?

pkg/printer/printer.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+}
+
+// ParamType prints:
+// foo(this <: <ParamType>):
+func (p *Printer) ParamType(param *sysl.Param) string {
+	if param.Type == nil {
+		return ""
+	}
+	// Ref.Appname.Part is the type name if the type is in the same package and the application name of where it's st
```
cuminandpaprika, line 238:
```suggestion
	// Ref.Appname.Part is the type name if the type is in the same package and the application name of where it's at
```

pkg/printer/tests/sequence.sysl
```diff
@@ -0,0 +1,6 @@
+Server:
+    !type Foo:
+        foo <: sequence of Bar
+
+    !type Bar:
+        bar <: sequence of string
```
cuminandpaprika, line 6: Newline

pkg/printer/alphabetical.go
joshcarp: Encountered some strange behaviour where sometimes there were empty endpoint names

pkg/printer/README.md
joshcarp: 👍

pkg/printer/alphabetical.go
cuminandpaprika: Think we can do the check for all of the above^ Helps make it consistent at least, so that all empty appnames param names etc are ignored?

PR #828, opened 2020-05-01
Add catalog command

joshcarp:
Calls out to github.com/anz-bank/sysl-catalog


cuminandpaprika:
Looks good, will do another review when CI issue are fixed up, think you can add the relevant docs to Sysl Commands docs/website/content/docs/commands and Sysl Features in docs/website/content/features

PR #829, opened 2020-05-01
Adds support for OpenAPI3 export from sysl

cuminandpaprika:
This PR adds an OpenAPI3 exporter which offers partial support for the OpenAPI3 Spec. 

Checkout this branch and test it using 
`dist/sysl export -f openapi3 pkg/exporter/test-data/openapi3/petstore.sysl  `

Currently, supports conversion of sysl endpoints, parameters(header,body,path,query), response types and type definitions.
Also supports population of the Info and Server objects from defined attributes in the sysl file.

Limitations include:
- Only `application/json` is supported
- No support for Discriminator Objects https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#discriminatorObject
- No support for Security Objects https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#security-scheme-object
- No support for the License Object https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#licenseObject
- No support for Callbacks https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#callbackObject
- No support for Links https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#linkObject

The majority of these limitations are due to the lack of corresponding representations in Sysl.

- Add openapi3 support to exporter
- Allow swagger or openapi2 to be used for mode flags
- Fix issue with query parameter reference parsing
- Add details to language spec in regards to parameters and return types

Checklist:
- [x] Added related tests


pkg/exporter/openapi3_test.go
```diff
@@ -0,0 +1,165 @@
...
+	exporter := MakeOpenAPI3Exporter(simpleApps, &logrus.Logger{})
+	err := exporter.Export()
+	assert.NoError(t, err)
+	limitRequired := exporter.openapi3["TestApp"].Paths["/pets"].Get.Parameters.GetByInAndName("query", "limit").Required
+	assert.False(t, limitRequired)
+}
+
+func TestMapEnums(t *testing.T) {
+	simpleApps := map[string]*syslwrapper.App{
+		"TestApp": {
```
joshcarp, line 123: I like the idea of this, but I think it might be a little cleaner if you could just make an app/module directly from a string; currently you need to pass a file system in

docs/website/content/docs/language.md
```diff
@@ -37,6 +37,49 @@ Notes about sysl syntax:
   - All lines after `:` should be indented. The only exception to this rule is when you want to use the shortcut `...`.
   - The `...` (aka shortcut) means that we don't have enough details yet to describe how this endpoint behaves. Sysl allows you to take an iterative approach in documenting the behaviour. You add more as you know more.
 
+### Application Names
```
joshcarp, line 40: This isn't openAPI?

pkg/exporter/openapi3.go
```diff
@@ -0,0 +1,223 @@
...
+			operation.AddResponse(parseResponseCode(value.Name), response)
+		}
+
+		spec.AddOperation(path, method, operation)
+	}
+
+	// err := spec.Validate(context.Background())
+	// if err != nil {
+	// 	panic(err)
+	// }
```
joshcarp, line 138: comment

cuminandpaprika: Thanks, will clean this up

docs/website/content/docs/language.md
cuminandpaprika: Ahh, yeah some of these docs are related to how to define an application in Sysl so that they can be exported to Openapi3. Happy to move this to another PR if needed

pkg/exporter/openapi3_test.go
cuminandpaprika: Sounds good, will make a utility function for that in a future PR

cuminandpaprika:
> Lots of files there that don't look to be openapi related

Fixed up the comment, there's a bit of moving of test files around, and a bugfix in the sysl parser for path parameters, should be a lot clearer if you review commit by commit, rather than all changes at once (sorry should have put that in the PR description)

PR #831, opened 2020-05-04
Fixes an issue with set and table support in syslwrapper

cuminandpaprika:
Changes proposed:

- Add support for set in responses and type definitions
- Add basic support for relation type in sysl type definitions

Checklist:
- [x] Added related tests

### DIFF
--- a/pkg/syslwrapper/app.go
+++ b/pkg/syslwrapper/app.go
@@ -165,14 +165,14 @@ func (am *AppMapper) mapResponse(stmt []*sysl.Statement, appName string) map[str
 			Type: returnType,
 		}
 	}
-
 	return responseTypes
 }
 
 // Checks if the return value is a complex type such as sequence of string
 func (am *AppMapper) mapReturnType(retValue string, appName string) *Type {
 	var returnType *Type
-	if strings.Contains(retValue, "sequence of ") {
+	switch {
+	case strings.Contains(retValue, "sequence of "):
 		listType := strings.Replace(retValue, "sequence of ", "", 1)
 		returnType = &Type{
 			Type: "list",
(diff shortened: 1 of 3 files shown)


pkg/syslwrapper/app.go
```diff
@@ -96,7 +96,9 @@ func (am *AppMapper) IndexTypes() map[string]*sysl.Type {
 func (am *AppMapper) ConvertTypes() map[string]*Type {
 	simpleTypes := make(map[string]*Type)
 	for typeName, syslType := range am.Types {
-		simpleTypes[typeName] = am.MapType(syslType)
+		if am.MapType(syslType) != nil {
+			simpleTypes[typeName] = am.MapType(syslType)
```
ashwinsajiv, line 100: Can this -am.MapType(syslType) be stored in a meaningful name? So can be made like -
if meaningfulName := am.MapType(syslType); meaningfulName != nil {
        simpleTypes[typeName] = meaningfulName
}

pkg/syslwrapper/app.go
```diff
@@ -121,7 +123,9 @@ func (am *AppMapper) mapAttributes(attributes map[string]*sysl.Attribute) map[st
 func (am *AppMapper) mapTypes(appName string, syslTypes map[string]*sysl.Type) map[string]*Type {
 	simpleTypes := make(map[string]*Type, 15)
 	for typeName := range syslTypes {
-		simpleTypes[typeName] = am.MapType(am.Types[appName+":"+typeName])
+		if am.MapType(am.Types[appName+":"+typeName]) != nil {
+			simpleTypes[typeName] = am.MapType(am.Types[appName+":"+typeName])
```
ashwinsajiv, line 127: Similar changes to this

pkg/syslwrapper/app.go
cuminandpaprika: Yup, agreed, can definitely do that

pkg/syslwrapper/app.go
cuminandpaprika: No worries, will do!

cuminandpaprika: Fixed up in https://github.com/anz-bank/sysl/pull/831/commits/49670ff4551670504b4e47b555a6e4b429979239

pkg/syslwrapper/app.go
cuminandpaprika: Fixed up in https://github.com/anz-bank/sysl/pull/831/commits/49670ff4551670504b4e47b555a6e4b429979239

PR #837, opened 2020-05-06
add go mod init check before go get

ChloePlanet:
- check go.mod before go get
- reverse the go get direction
- add `-u` to `go get`

Checklist:
- [x] Added related tests


pkg/mod/fs_test.go
```diff
@@ -38,68 +40,84 @@ func TestOpenLocalFile(t *testing.T) {
...
+	_, err := fs.Create("go.mod")
+	assert.NoError(t, err)
+	defer removeFile(t, fs, "go.mod")
 
 	filename := "github.com/anz-bank/sysl/demo/examples/Modules/deps.sysl"
 	_, memfs := syslutil.WriteToMemOverlayFs("/")
-	fs := NewFs(memfs, "")
-	f, err := fs.Open(filename)
+	mfs := NewFs(memfs, "")
+	f, err := mfs.Open(filename)
```
marcelocantos, line 63: nit: There's a lot of common structure between these tests. It might be worth seeing if there's a clean refactoring.

pkg/mod/gomod.go
```diff
@@ -22,23 +21,26 @@ type goModule struct {
 }
 
 func goGetByFilepath(filename string) error {
-	dir := filepath.Dir(filename)
-	re := regexp.MustCompile(`^\.|\.\.|/$`)
+	names := strings.Split(filename, "/")
+	if len(names) > 0 {
```
marcelocantos, line 25: nit: Fold into previous line.

pkg/mod/gomod.go
```diff
@@ -22,23 +21,26 @@ type goModule struct {
...
+		gogetPath := names[0]
+
+		for i := 1; i < len(names); i++ {
```
marcelocantos, line 28:
```suggestion
		for _, name := range names[1:] {
```

pkg/mod/gomod.go
```diff
@@ -22,23 +21,26 @@ type goModule struct {
...
+			if err == nil {
+				return nil
+			}
+			logrus.Debugf("go get %s error: %s\n", gogetPath, err.Error())
 
-	for !re.MatchString(dir) {
-		err := goGet(dir)
-		if err == nil {
-			return nil
+			gogetPath = path.Join(gogetPath, names[i])
```
marcelocantos, line 35:
```suggestion
			gogetPath = path.Join(gogetPath, name)
```

marcelocantos, line 35: Alternatively, slightly less efficient, but cleaner (And who cares about efficiency when you're hitting the filesystem? :smile:):
```go
for i = range names[1:] {
	err := goGet(path.Join(names[:1+i]...))
	if err == nil {
		return nil
	}
	logrus.Debugf("go get %s error: %s\n", gogetPath, err.Error())
}
```

PR #848, opened 2020-05-10
Simplify sysl printer

joshcarp:
* Removes unnecessary struct
* Simplifies names

### DIFF
--- a/pkg/printer/printer.go
+++ b/pkg/printer/printer.go
@@ -4,11 +4,11 @@ package printer
 
 import (
 	"fmt"
+	"github.com/anz-bank/sysl/pkg/syslutil"
 	"io"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/printer/printer.go
```diff
@@ -18,215 +18,222 @@ const (
...
 	}
 	patterns := GetPatterns(attrs)
-	if len(patterns) > 0 {
-		fmt.Fprint(p.Writer, "[")
-		for i, pattern := range patterns {
-			fmt.Fprintf(p.Writer, "~%s", pattern)
-			if i != len(patterns)-1 {
-				fmt.Fprintf(p.Writer, ", ")
-			}
+	numPatterns := len(patterns)
```
anzdaddy, line 119: nit: Not strictly necessary. `len(patterns)` is practically a NOOP.

PR #856, opened 2020-05-18
default to openapi when importing yaml/json and handle swagger files …

anz-gordonj7:
…via openapi importer

```
import spec.yaml # originally assumed it was swagger, would silently fail if it was openapi
import spec.yaml ~openapi3 # this forced it to openapi and worked

```

This PR changes the import handling so any yaml which isnt explicitly moded will be loaded via the OpenAPI handler which will offload a swagger doc to the swagger loader.


ashwinsajiv:
LGTM

PR #866, opened 2020-05-25
Refactorforcatalog

ashwinsajiv:
Data model diagram for mermaid for a specific type


pkg/mermaid/datamodeldiagram/datamodeldiagram.go
```diff
@@ -10,6 +10,10 @@ import (
 
 //cause linter..
 const ref = "ref"
+const enum = "enum"
+const mapS = "map"
+const relation = "relation"
+const tuple = "tuple"
```
cuminandpaprika, line 16: These could be declared in a const block, might be a bit cleaner

const (
   ref         = "ref"
   enum    = "enum"
   mapS    = "map"
   relation = "relation"
   tuple     = "tuple"
)

pkg/mermaid/datamodeldiagram/datamodeldiagram.go
```diff
@@ -10,6 +10,10 @@ import (
 
 //cause linter..
```
cuminandpaprika, line 11:
```suggestion
// Remove goconst linter warnings
```

pkg/mermaid/datamodeldiagram/datamodeldiagram.go
ashwinsajiv: Addressed!

pkg/mermaid/datamodeldiagram/datamodeldiagram.go
ashwinsajiv: Addressed!

pkg/mermaid/datamodeldiagram/datamodeldiagramwithappandtype.go
```diff
@@ -0,0 +1,81 @@
...
+	mapper.ConvertTypes()
+	return generateDataModelDiagramWithAppAndTypeHelper(mapper.SimpleTypes, appName, typeName,
+		&[]externalLink{}, &[]string{}, &[]externalLink{})
+}
+
+func generateDataModelDiagramWithAppAndTypeHelper(simpleTypes map[string]*syslwrapper.Type, appName string,
+	typeName string, externalLinks *[]externalLink, appTypes *[]string, eLinks *[]externalLink) (string, error) {
+	var result string
+	appType1 := fmt.Sprintf("%s:%s", appName, typeName)
+	result = "%% AUTOGENERATED CODE -- DO NOT EDIT!\n\nclassDiagram\n"
```
cuminandpaprika, line 23: `%% AUTOGENERATED CODE -- DO NOT EDIT!\n\nclassDiagram\n` is also defined in datamodeldiagram.go, perhaps it'd be defined as a single constant, that can be shared?

I think it'd also be cleaner to give them names like so
```
const generatedHeader =""%% AUTOGENERATED CODE -- DO NOT EDIT!\n" 
const classDiagramHeader ="\nclassDiagram\n"
```

ashwinsajiv: Addressed!

pkg/mermaid/datamodeldiagram/datamodeldiagramwithappandtype.go
```diff
@@ -0,0 +1,81 @@
...
+		result += fmt.Sprintf(" class %s {\n", mermaid.CleanString(appType))
+		switch typeValue.Type {
+		case relation, tuple:
+			result += printProperties(typeValue.Properties, appType, externalLinks)
+		case enum:
+			result += printEnum(typeValue.Enum)
+		case mapS:
+			result += printMap(typeValue.Properties, appType, typeValue.PrimaryKey, externalLinks)
+		default:
+			result += ""
```
cuminandpaprika, line 48: Maybe logging a warning here would be better? `result +=""` doesn't do anything right?

pkg/mermaid/integrationdiagram/integrationdiagram.go
```diff
@@ -15,12 +15,30 @@ type integrationPair struct {
...
+//The resulting string is the mermaid code for the integration diagram
 func GenerateIntegrationDiagram(m *sysl.Module, appName string) (string, error) {
+	if appName == "" {
+		return generateEntireIntegrationDiagramHelper(m, &[]integrationPair{})
+	}
 	return generateIntegrationDiagramHelper(m, appName, &[]integrationPair{}, true)
 }
 
-//generateIntegrationDiagramHelper is a helper which has additional arguments which need not be entered by the user
+//generateEntireIntegrationDiagramHelper is a helper which is generates an entire integration diagram
```
cuminandpaprika, line 26: I think the function name is a bit confusing, at least I can't seem to fully grasp what it does from the name.

Does it general an integration diagram for ALL applications?
Maybe `generateFullDiagram` or `generateAll`

pkg/mermaid/integrationdiagram/integrationdiagram.go
```diff
@@ -15,12 +15,30 @@ type integrationPair struct {
 }
 
 //GenerateIntegrationDiagram accepts an application name as input and returns a string (and an error if any)
-//The resulting string is the mermaid code for the integration diagram for that application
+//The resulting string is the mermaid code for the integration diagram
```
cuminandpaprika, line 18: Just looking for some clarification for this function, what's the reason behind wrapping `generateEntireIntegrationDiagramHelper ` and `generateIntegrationDiagramHelper ` in a single function here instead of having 
`GenerateEntireIntegrationDiagram` and `GenerateIntegrationDiagram` as two public functions?

It seems to me like it'd be easier for users to use whichever one is appropriate, rather than needing to know that passing an empty string into the appName param will generate a diagram for everything

pkg/mermaid/datamodeldiagram/datamodeldiagramwithappandtype.go
ashwinsajiv: the class diagram header isn't repeated, hence I haven't changed it.

pkg/mermaid/datamodeldiagram/datamodeldiagramwithappandtype.go
ashwinsajiv: Yes! I will added a logger once I get the initial implementation working on catalog. I have a tracker for that issue.

pkg/mermaid/integrationdiagram/integrationdiagram.go
ashwinsajiv: Yep, makes sense. I'll split them into two functions.

pkg/mermaid/integrationdiagram/integrationdiagram.go
ashwinsajiv: Yep, I'll change that!

ashwinsajiv:
> Looks really good overall, you've split up the functions really well so they're nice and small. Just a few small nitpicks about duplication of strings between `datamodeldiagram.go` and `datamodeldiagramwithappandtype.go`
> 
> Also, have you added some new tests for `datamodeldiagramwithappandtype.go`?
> 
> One last nitpick: Could we possibly use something a bit shorter than `datamodeldiagramwithappandtype.go`?
> 
> I actually wouldn't mind seeing datamodeldiagramwithappandtype moved into `datamodeldiagram.go` since it's handling the same concern: generating datamodel diagrams. If you needed to change some behaviour in the future, you'd want to make changes to both files. Also that way, it's easier to see any duplications, and you can put for example `GenerateDataModelDiagramWithAppAndType` right below `GenerateDataModelDiagram`. Thoughts?

Yeah can do that, there aren't going to be any duplication but the feature itself was kinda getting bigger and I wanted to split it. I'll put them together if that is more readable.

pkg/mermaid/datamodeldiagram/datamodeldiagramwithappandtype.go
cuminandpaprika: 👍

pkg/mermaid/datamodeldiagram/datamodeldiagram.go
```diff
@@ -8,44 +8,101 @@ import (
...
+		&[]externalLink{}, &[]string{}, &[]externalLink{})
+}
+
+//generateDataDiagramWithAppAndTypeHelper is a helper which has additional arguments
+func generateDataDiagramWithAppAndTypeHelper(simpleTypes map[string]*syslwrapper.Type, appName string,
+	typeName string, externalLinks *[]externalLink, appTypes *[]string, eLinks *[]externalLink) (string, error) {
+	var result string
+	appType1 := fmt.Sprintf("%s:%s", appName, typeName)
+	result = mermaid.GeneratedHeader + "classDiagram\n"
+	for appType, value := range simpleTypes {
```
cuminandpaprika, line 49: Looks like this for loop is duplicated at line 89? (I might've missed a difference). Could this be pulled out into a function, that gets reused between generateDataDiagramWithAppAndTypeHelper and generateFullDataDiagramHelper?

ashwinsajiv: Just a tiny difference, I reused the same to get all the possible links between types. That one doesn't actually build the string.

PR #881, opened 2020-06-09
Fix the `sysl export` command line to be more user friendly

anz-gordonj7:
New behaviour:
If an app is supplied on the command line then only it will be exported,
otherwise all apps defined will be exported.

If the filename doesnt include the $(appname) magic the appname will be infixed in the filename

### DIFF
--- a/cmd/sysl/cmd_export.go
+++ b/cmd/sysl/cmd_export.go
@@ -118,19 +118,30 @@ func (p *exportCmd) Execute(args cmdutils.ExecuteArgs) error {
 		return err
 	}
 
-	if strings.Contains(p.out, "%(appname)") {
-		for appName, syslApp := range args.Modules[0].GetApps() {
+	var writeCount int
+	for appName, syslApp := range args.Modules[0].GetApps() {
+		if appName == p.appName || p.appName == "" {
 			outputFileName := cmdutils.MakeFormatParser(p.out).LabelApp(appName, "", syslApp.GetAttrs())
+			if err := args.Filesystem.MkdirAll(filepath.Dir(outputFileName), os.ModePerm); err != nil {
+				return err
+			}
+			if p.appName == "" && outputFileName == p.out {
+				ext := filepath.Ext(outputFileName)
+				// convert out.yaml something like out.Fooapp.yaml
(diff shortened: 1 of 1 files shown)


cmd/sysl/cmd_export.go
```diff
@@ -118,19 +118,30 @@ func (p *exportCmd) Execute(args cmdutils.ExecuteArgs) error {
...
+	for appName, syslApp := range args.Modules[0].GetApps() {
+		if appName == p.appName || p.appName == "" {
 			outputFileName := cmdutils.MakeFormatParser(p.out).LabelApp(appName, "", syslApp.GetAttrs())
+			if err := args.Filesystem.MkdirAll(filepath.Dir(outputFileName), os.ModePerm); err != nil {
+				return err
+			}
+			if p.appName == "" && outputFileName == p.out {
+				ext := filepath.Ext(outputFileName)
+				// convert out.yaml something like out.Fooapp.yaml
+				outputFileName = fmt.Sprintf("%s.%s%s", strings.TrimSuffix(outputFileName, ext), appName, ext)
```
cuminandpaprika, line 131: Interesting that you've gone with output file names as `out.Fooapp.yaml` rather than just `Fooapp.yaml` Typically I've seen OpenAPI specs named the same as the service so I'd expect users to want to export as AppName.yaml etc. 

In this case, if I were to use `sysl export myapp.sysl -o specs` would the output files be named .FooApp.yaml?

anz-gordonj7: It was more of a "save the user from himself" thing. This printf is only used when no app was specified, so in your example you'd end up with `specs.FooApp.yaml`, but of course you really should be using the %(appname) magic in the filename.

anz-gordonj7: I chose to mash the app name into the specified filename instead of using the appname because the user may actually have a appname.yaml they didnt mean to overwrite.

cuminandpaprika: Okay that's a pretty good reason actually, makes sense! 
Something I've taken to using is appending my openapi files with FooApp.oas.yml or FooApp.oas.json so that it's obvious that these are openapi spec files. 

Could be a nice way of encouraging users to use this convention here.

PR #894, opened 2020-06-17
Update install rule in Makefile per https://sysl.io/docs/install/

orlade-anz:
- Change make install to do what the docs suggest (current behaviour requires sudo)

Checklist:
- [ ] Added related tests


orlade-anz:
This is the wrong solution; we need the output of `go build` since it has the version info.

PR #905, opened 2020-06-18
Add missing details to setup docs

cuminandpaprika:
Some minor documentation improvements to add missing details around building Sysl based on recent questions asked by new sysl users.

- Add minimum version of golangci-lint needed
- Add requirement for GOPATH env var
- Add comment to makefile clarifying make install target

Checklist:
- [x] Added related tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ buildlsp: ## Build sysllsp into the ./dist folder
 .PHONY: release
 release: $(PLATFORMS) ## Build release binaries for all supported platforms into ./release
 
-install: build ## Install the sysl binary into $(GOPATH)/bin
+install: build ## Install the sysl binary into $(GOPATH)/bin. We don't use go install because we need to pass in LDFLAGS.
 	cp ./dist/sysl $(GOPATH)/bin
 
 clean: ## Clean temp and build files
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@ buildlsp: ## Build sysllsp into the ./dist folder
 release: $(PLATFORMS) ## Build release binaries for all supported platforms into ./release
 
 install: build ## Install the sysl binary into $(GOPATH)/bin. We don't use go install because we need to pass in LDFLAGS.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -48,7 +48,8 @@ buildlsp: ## Build sysllsp into the ./dist folder
 .PHONY: release
 release: $(PLATFORMS) ## Build release binaries for all supported platforms into ./release
 
-install: build ## Install the sysl binary into $(GOPATH)/bin
+install: build ## Install the sysl binary into $(GOPATH)/bin. We don't use go install because we need to pass in LDFLAGS.
```
joshcarp, line 51: isn't gopath/bin only for binaries installed using `go get` or `go install`; any user built binaries should go in `/bin/` or `/usr/local/bin/`
Go doesn't know how to version binaries in `$GOPATH/bin`; a command like `go get` or `go install` might overwrite this binary and it won't have the build info you desire

orlade-anz: Good point, no idea.

cuminandpaprika: That's a fair point actually, usually people add the `GOPATH/bin` to the end of their PATH var in their `.bashrc` or `.zshrc` files so it would be lower priority.

So if we were to install it into one of those directories, whenever they try and update sysl using go get github.com/anz-bank/cmd/sysl, it wouldn't actually update it, since the version installed in /bin/ or /usr/local/bin would take precedence right?

So that's one argument for installing it into the `GOPATH/bin` instead? I think if it gets overwritten, that's kind of okay. Thoughts?

cuminandpaprika: Also, /bin/ and /usr/local/bin are *nix specific whereas `GOPATH/bin`works on windows?

orlade-anz: IMO overwrites are fine in this case. It's unlikely to get some kind of silent auto-update, and `sysl info` should still be correct for whatever is installed.

FWIW I added $GOPATH/bin to the front of my path, presumably following some instructions.

PR #910, opened 2020-06-22
fix #882, add better validator

### LINKED ISSUE #882: Make a better Validator
Currently sysl doesn't have a proper validator, which is meant to give information about whether the input sysl is correct or not. 
1. Making sure that calls actually exist: `Application <- Endpoint`; both Application and Endpoint should exist in the module
2. Warning about return types: `return type` should be a warning (only `return ok <: type` is "supported")
3. Warning when upper/lowercase redefinitions occur: `application` and `Application` should yield a warning

nofun97:
Fixes #882 .

`sysl validate` command now has linter warnings, it checks the following:

1. Return statements that does not use `ok`
2. Check that endpoint exists in a call statement
3. Check for case-sensitive app definition

Checklist:
- [x] Added related tests


pkg/parse/listener_impl.go
```diff
@@ -1009,10 +1250,17 @@ func (s *TreeShapeListener) ExitHttp_path(*parser.Http_pathContext) {
 
 // EnterRet_stmt is called when production ret_stmt is entered.
 func (s *TreeShapeListener) EnterRet_stmt(ctx *parser.Ret_stmtContext) {
+	payload := strings.Trim(ctx.TEXT().GetText(), " ")
+	if s.lintMode && !regexp.MustCompile(`<:`).MatchString(payload) {
```
anz-gordonj7, line 1254: this regex will not allow `return ok` or `return error`

pkg/parse/tests/invalid_call_self_method.sysl
```diff
@@ -0,0 +1,4 @@
+App:
+    /hi:
+        GET:
+            . <- POST /hi
```
cuminandpaprika, line 4: I had no idea the dot operator existed for sysl!

pkg/parse/lint_docs.md
```diff
@@ -0,0 +1,84 @@
...
+Currently, the linter can detect the following things:
+
+1. [Checks if return statements are not in the form of `return ok <: some_type`.](#return-statement-lint)
+2. [Checks if there are case-sensitive application definitions e.g. `App` and `app`.](#case-sensitive-applications-redefinition)
+3. [Checks if a call statement calls to an existing application, endpoint, and method.](#call-statement-linting)
+
+## Return Statement Lint
+
+For this feature, it is easier to lint during parsing. A simple regex match in
+the `EnterRet_stmt` function is done to check if it uses the `ok <: some_type`
```
cuminandpaprika, line 18:
```suggestion
the `EnterRet_stmt` function is done to ensure it uses the `ok <: some_type`
```

pkg/parse/lint_docs.md
```diff
@@ -0,0 +1,84 @@
+# Linter Design
```
cuminandpaprika, line 1: I'm delighted to see there's design docs! This is amazing!!

pkg/parse/listener_impl.go
nofun97: yep that's the idea. It only warns users so it is still allowed, there's just going to be a linter warning. https://github.com/anz-bank/sysl/issues/882#issue-636695948

pkg/parse/listener_impl.go
```diff
@@ -82,10 +86,247 @@ func NewTreeShapeListener() *TreeShapeListener {
...
+				continue
+			}
+			appNotExistLog(location, appName, call)
+		}
+	}
+
+	for appName, appData := range *s.linter.calls {
+		for endpoint, endpointData := range *appData.rec {
+			// if method maps == nil, it is not REST endpoint
+			if endpointData.rec == nil {
```
cuminandpaprika: Nitpick: Could potentially pull this out into a function for readability, but it's pretty readable as is, so just presenting this as an option
```
if isNotRESTEndpoint {
}

func isNotRESTEndpoint(rec Record) bool {
    return endpointData.rec == nil
}
```

pkg/parse/listener_impl.go
```diff
@@ -82,10 +86,247 @@ func NewTreeShapeListener() *TreeShapeListener {
...
+	lint := func(appName, endpoint, method, call string, locations map[string]bool) {
+		for location := range locations {
+			apps, exists := s.linter.apps[strings.ToLower(appName)]
+			if !exists {
+				appNotExistLog(location, appName, call)
+				continue
+			}
+			if app, exists := (*apps)[appName]; exists {
+				if endpoints, exists := (*app.rec)[endpoint]; exists {
+					// if method is empty string, it is linting simple endpoint
```
cuminandpaprika:
```suggestion
					// if method is empty string, it is linting simple endpoint, not a REST endpoint
```

nofun97:
@anz-gordonj7 
>     1. Move the linter functions to (at least) their own file, listener_impl.go is way too big already.

👍 will move it to its own file
  
>     2. why make these linter warnings instead of validation errors?
The plan is later to make `sysl validate` fails too but it seems that's going to be a big change. The sysl modules are loaded before executing the command so it's going to be tricky. So I thought this PR should be just about detecting the problems and have another PR that will make `sysl validate` fail.

pkg/parse/parse.go
```diff
@@ -266,15 +268,17 @@ func checkCalls(mod *sysl.Module, appname string, epname string, dst *sysl.State
 	case *sysl.Statement_Call:
 		app := syslutil.GetApp(s.Call.Target, mod)
 		if app == nil {
-			logrus.Warnf("%s::%s calls non-existent App: %s",
-				appname, epname, s.Call.Target.Part)
+			//TODO: delete this? this removes non-existent app call lint
```
cuminandpaprika, line 271: Can this be removed now that we handle linting of non existent apps elsewhere?
Same for the commented out code below

pkg/parse/listener_impl.go
nofun97: @anz-gordonj7 I just realised what you meant 😅, I'll try to change it.

pkg/parse/listener_impl.go
```diff
@@ -82,10 +86,247 @@ func NewTreeShapeListener() *TreeShapeListener {
...
+func (g *graph) recordApp(appName, location string) error {
+	if app, exists := (*g)[appName]; !exists {
+		(*g)[appName] = &graphData{newLocation(location), newAppEndpointGraph()}
+		return nil
+	} else if !app.locations[location] {
+		//TODO: find better way, to handle collector
+		app.locations[location] = true
+		return nil
+	}
+	return fmt.Errorf("recordApp: app already exists: %s %s", location, appName)
```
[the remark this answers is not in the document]
nofun97: I think it was just meant to handle in case it already exists in the graph, it really shouldn't happen and if it does, that means something is wrong in the implementation.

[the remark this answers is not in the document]
nofun97: I'll add that in the design docs

pkg/parse/listener_impl.go
nofun97: it's meant to be only used at a certain level of the graph, I think it would create confusion if I make a function for it.

pkg/parse/listener_impl.go
nofun97: fixed this

pkg/parse/lint_docs.md
```diff
@@ -0,0 +1,84 @@
...
+field's job is to collect data that are required to do its linting operation.
+The linting operation can happen during walking the abstract syntax tree or
+after that process. This depends on what the linter is detecting.
+
+Currently, the linter can detect the following things:
+
+1. [Checks if return statements are not in the form of `return ok <: some_type`.](#return-statement-lint)
+2. [Checks if there are case-sensitive application definitions e.g. `App` and `app`.](#case-sensitive-applications-redefinition)
+3. [Checks if a call statement calls to an existing application, endpoint, and method.](#call-statement-linting)
+
```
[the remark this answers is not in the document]
nofun97: added

PR #920, opened 2020-06-26
use http://www.plantuml.com/plantuml as default value

### LINKED ISSUE #919: Add default value when SYSL_PLANTUML is unset
### Current Sysl Version
v0.123.0

### Use-cases
`export SYSL_PLANTUML=http://www.plantuml.com/plantuml` is required to generate diagram at present. 
Adding a default value to it if SYSL_PLANTUML is unset.

ChloePlanet:
Fixes #919 .

- use http://www.plantuml.com/plantuml as default value

Checklist:

### DIFF
--- a/pkg/diagrams/plantuml.go
+++ b/pkg/diagrams/plantuml.go
@@ -16,7 +16,7 @@ import (
 
 const (
 	PlantUMLEnvVar  = "SYSL_PLANTUML"
-	PlantUMLDefault = "http://localhost:8080/plantuml"
+	PlantUMLDefault = "http://www.plantuml.com/plantuml" // "http://localhost:8080/plantuml"
 	testDir         = "../../tests/"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/diagrams/plantuml.go
```diff
@@ -16,7 +16,7 @@ import (
 
 const (
 	PlantUMLEnvVar  = "SYSL_PLANTUML"
-	PlantUMLDefault = "http://localhost:8080/plantuml"
+	PlantUMLDefault = "http://www.plantuml.com/plantuml" // "http://localhost:8080/plantuml"
```
anzdaddy, line 19: This is a dangerous default, since it will result in accidental leakage of work product from organisations using Sysl internally. A better solution is to keep the default but provide helpful instructions in the error output if localhost:8080 doesn't work.

PR #933, opened 2020-07-03
Feature/simpler openapi importing

anz-gordonj7:
Rewrites the swagger/openapi importer to be much simpler! will automatically convert importer swagger2 -> openapi3 references correctly.

This version creates far fewer alias -> string types which should improve the sysl clarity (may cause some breakages but should be minor).

edit: The openapi.go change is basically a re-write, dont bother looking at the diff, just look at the new file version


pkg/importer/openapi.go
```diff
@@ -3,105 +3,150 @@ package importer
...
-	types         TypeList
-	// intermediateTypes is a temporary list which places the type is in parsing process still.
-	// It can help to support circular dependency, like type A has an array contains type A itself.
-	intermediateTypes TypeList
-	swaggerRoot       string
-	globalParams      Parameters
-}
-
-func (l *OpenAPI3Importer) Load(input string) (string, error) {
+func NewOpenAPILoader(logger *logrus.Logger, fs afero.Fs) *openapi3.SwaggerLoader {
```
cuminandpaprika, line 28: nit: it seems NewOpenAPILoader isn't used outside the `import` package, should it be public or should it be made private to encourage the use of the `NewOpenAPIV3Importer` API

pkg/importer/swagger.go
```diff
@@ -83,6 +80,23 @@ func convertToOpenAPI3(data []byte) (*openapi3.Swagger, string, error) {
...
+			"#/parameters/", "#/components/parameters/",
+		)
+		oa3replaced := replacer.Replace(string(oa3text))
+		_ = json.Unmarshal([]byte(oa3replaced), &openapiv3) //nolint: errcheck // we dont care about the error here
+	}
+
+	err = loader.ResolveRefsIn(openapiv3, uri)
+	if err != nil {
+		return nil, "", err
+	}
```
nofun97, line 99:
```suggestion
	if err = loader.ResolveRefsIn(openapiv3, uri); err != nil {
		return nil, "", err
	}
```

pkg/importer/tests-openapi/nested-types.sysl
```diff
@@ -21,38 +21,29 @@ testapp "Fruit API" [package="package_foo"]:
             @json_tag = "veggies"
 
     !type Details:
-        month <: Details_month?:
+        month <: string?:
```
nofun97, line 24: probably out of scope but looking at the openapi spec, it seems like this is supposed to be an enum https://github.com/anz-bank/sysl/blob/master/pkg/importer/tests-openapi/nested-types.yaml#L59

pkg/importer/tests-openapi/nested-types.sysl
```diff
@@ -21,38 +21,29 @@ testapp "Fruit API" [package="package_foo"]:
...
             @json_tag = "transDet"
 
-    !alias Details_month:
-        string
-
     !type Fruit:
         details <: sequence of Details?:
             @json_tag = "details"
-        entity <: Fruit_entity?:
+        entity <: string?:
```
nofun97, line 34: ditto

pkg/importer/tests-openapi/nested-types.sysl
```diff
@@ -21,38 +21,29 @@ testapp "Fruit API" [package="package_foo"]:
...
             @json_tag = "name"
 
-    !alias Fruit_entity:
-        string
-
     !type Veggies:
         details <: sequence of Details?:
             @json_tag = "details"
-        entity <: Veggies_entity?:
+        entity <: string?:
```
nofun97, line 44: ditto

pkg/importer/tests-swagger/enums.sysl
```diff
@@ -15,12 +15,9 @@ testapp "foo" [package="package_foo"]:
     !type Attachment:
         data <: string:
             @json_tag = "data"
-        type <: Attachment_type:
+        type <: string:
```
nofun97, line 18: ditto

pkg/importer/tests-swagger/overwrite-header.sysl
```diff
@@ -12,12 +12,9 @@ testapp "Sample API" [package="package_foo"]:
         | API description in Markdown.
 
     /users/{id<:int}:
-        GET (metadata <: _param_metadata [~header, ~required, name="metadata"]):
+        GET (metadata <: string [~header, ~required, name="metadata"]):
```
nofun97, line 15: ditto

pkg/importer/tests-swagger/param-array.sysl
```diff
@@ -9,7 +9,7 @@ testapp "Simple" [package="package_foo"]:
         | No description.
 
     /simple:
-        PUT (DocumentRequest <: Document [mediatype="application/json", ~body]):
+        PUT (DocumentRequest <: DocumentRequest [mediatype="application/json", ~body]):
```
nofun97, line 12: nit: the alias name probably shouldn't be the same as the param name

pkg/importer/tests-swagger/path-var-overwrite.sysl
```diff
@@ -12,7 +12,7 @@ testapp "Sample API" [package="package_foo"]:
         | API description in Markdown.
 
     /users/{id<:int}:
-        GET (metadata <: _param_metadata [~header, ~required, name="metadata"]):
+        GET (metadata <: string [~header, ~required, name="metadata"]):
```
nofun97, line 15: this should be `bool` https://github.com/anz-bank/sysl/blob/master/pkg/importer/tests-swagger/path-var-overwrite.yaml#L16

pkg/importer/openapi.go
```diff
@@ -3,105 +3,150 @@ package importer
...
-		AppName: l.appName,
-		Package: l.pkg,
-	}, l.basePath), l.types, endpoints...); err != nil {
-		return "", err
-	}
-	return result.String(), nil
+func (o *openapiv3) WithPackage(packageName string) Importer {
+	o.pkg = packageName
+	return o
+}
```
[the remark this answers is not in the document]
anz-gordonj7: I agree, I wasnt the one that added the interface this is matching

pkg/importer/tests-openapi/nested-types.sysl
anz-gordonj7: correct, there is a `if 0`'d out block of code which I'll reenable very soon which adds sysl attributes to these strings to add the legal values

nofun97: there are enums in in sysl https://github.com/anz-bank/sysl/blob/master/pkg/parse/tests/enum.sysl but looking at the enums in the openapi spec, maybe sysl enum won't work as well (enums in sysl uses numbers).

PR #935, opened 2020-07-05
Revert "Refactor Endpoints (#922)"

cuminandpaprika:
Revert https://github.com/anz-bank/sysl/pull/922 as it breaks sysl transforms as they expect an `error` type rather than an error code

This reverts commit 46abf49699f26f4437ddc3e80de0531dc6842b65.


pkg/importer/openapi.go
```diff
@@ -404,11 +396,15 @@ func isSchemaDefinedObject(ref *openapi3.SchemaRef) bool {
...
 	var responses []Response
-	typePrefix := getSyslSafeName(cleanEndpointPath(path)) + "_"
+	typePrefix := strings.NewReplacer(
+		"/", "_",
+		"{", "_",
+		"}", "_",
+		"-", "_").Replace(path) + "_"
 	for statusCode, resp := range op.Responses {
-		text := statusCode
+		text := "error"
```
anz-gordonj7, line 407: This line was probably the only one that needed to be put back...

PR #948, opened 2020-07-10
Fix multiple error types

cuminandpaprika:
This PR appends the error code to the type name, without changing the error response definition (which was previously a breaking change for transforms) This change is now hidden behind a toggle flag, so that it will not break arrai transforms

Behaviour with flag: https://github.com/anz-bank/sysl/pull/948/files#diff-96ec2bd3bb0448da4b0b23b92469582a
Behaviour without flag: https://github.com/anz-bank/sysl/pull/948/files#diff-ac6fdb35ec698c09e006ff0cb5a83a64

Changes:
- Fixes an importer issue with multiple error return types being duplicated as types with the same name.
- Also escapes output regex string, which was causing invalid Sysl to be output

Checklist:
- [x] Added related tests


pkg/importer/tests-openapi/multiple-return-values.sysl
```diff
@@ -5,28 +5,52 @@
...
+    @version = "1.0"
     @description =:
         | No description.
 
     /test:
         GET:
             | No description.
-            return error
-            return ok <: _test_ok
+            return error <: ErrorResponse
```
anz-gordonj7, line 15: this change will cause breakages with the arrai transforms

pkg/importer/importer_test.go
```diff
@@ -77,6 +96,15 @@ func TestLoadXSDFromTestFiles(t *testing.T) {
 	})
 }
 
+func TestImportWithFlags(t *testing.T) {
```
andrewemeryanz, line 99: Good. Please add a test that tests existing behaviour (without the flag set).

PR #982, opened 2020-07-30
Clean up some build and release scripts

orlade-anz:
This requires inspection by folks familiar with the build and release process to make sure it doesn't break anything, but I don't think it should.

This is not a complete refactor, but a step towards a simpler world of config. We should diligently delete everything we don't strictly need, or people who don't know what it does will be too afraid to change it.


Makefile
```diff
@@ -8,55 +11,45 @@ else
 	TESTEXE := go test ./...
 endif
 
-all: test lint build buildlsp coverage examples ## test, lint, build, coverage test run
+all: lint build buildlsp install examples golden test coverage
```
cuminandpaprika, line 14: I'm not sure that we should include golden to run on every make (This is because some of the golden.pb files are non deterministic, since google decided to put in random spaces to make it so. This will introduce quite a bit of noise in diffs as a result)

PR #996, opened 2020-08-07
Ingest Spanner SQL to produce Sysl specification.

springwiz:
Ingest Spanner DDL using wbnf grammar.

Fixes https://github.com/anzx/acceleration/issues/327.

1.  Define a wbnf grammar that describes Spanner DDL.
2. Define the transforms that ingest the sql file and output arrai native data types.
3. Write the output to a file.
4. Glue code written as go importer which calls arrai scripts created above.

Checklist:
- [x] Added related tests


pkg/importer/spanner/import
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env sh
+arrai r import.arrai $1 > $2
```
orlade-anz, line 2: Nit: this file doesn't appear to add any particular value, and worst case could encourage non-standard approaches to running arrai scripts. Is there some context in which simply writing `arrai run import foo > bar` is too inconvenient?

Feel free to have scripts like this locally, but don't include in the design of a feature or commit to Git.

springwiz: It wouldn't be there in the final PR probably, only serves as the convenience script currently.

springwiz: Also the fact that this would be integrated to sysl command line eventually, this script would just serve as a test ground until we actually integrate with sysl command line.

pkg/importer/spanner/test-spanner.arrai
```diff
@@ -0,0 +1 @@
... {(attribute: 'AccountNum', reference_attribute: 'AccountNum', reference_table: 'Account')}), (constraint_name: 'FK_CustomerID', foreign_keys: {(attribute: 'CustomerID', reference_attribute: 'CustomerID', reference_table: 'Customer')})}, name: 'CustomerHasAccount', primary_key: {'AccountNum', 'CustomerID'}), (attributes: {(array: {}, length: 10, name: 'Mobile', nullable: true, options: {}, type: 'string'), (array: {}, length: 36, name: 'CustomerID', nullable: {}, options: {}, type: 'string'), (array: {}, length: 64, name: 'FirstName', nullable: {}, options: {}, type: 'string'), (array: {}, length: 64, name: 'LastName', nullable: {}, options: {}, type: 'string'), (array: {}, length: 256, name: 'Email', nullable: true, options: {}, type: 'string')}, cluster: {}, foreign_keys: {}, name: 'Customer', primary_key: {'CustomerID'}), (interleaved_table: {}, key_part: 'AccountNum DESC', name: 'AccountsByNum', nullfiltered: {}, storing_col: {}, table_name: 'Account', unique: {}), (interleaved_table: 'Customer', key_part: 'Email ASC', name: 'CustomerByEmail', nullfiltered: 'NULL_FILTERED', storing_col: ['Email', 'Mobile'], table_name: 'Customer', unique: 'UNIQUE'), (name: 'CustomerAccounts')})
```
orlade-anz, line 1: Nit: If this is what the output is expected to look like (i.e. it's a golden file), please call it `*.golden.arrai` so it's apparent, and ideally generate it with `//fmt.pretty` so it's more human readable. The comparison should be done in arr.ai so the whitespace shouldn't matter.

pkg/importer/spanner/test-spanner.sql
```diff
@@ -0,0 +1,37 @@
+CREATE DATABASE CustomerAccounts;
```
orlade-anz, line 1: Nit: Can you add a comment at the top of this file, partly to describe what it is, and partly to test that the grammar can handle comments?

springwiz: ok

pkg/importer/spanner/test-spanner.arrai
springwiz: I don't intend to make it a golden file, instead there will be a golden file created for sysl output generated from this file. This file may not even exist once we get the sysl output generated.

springwiz: I don't intend to make it a golden file, instead there will be a golden file created for sysl output generated from this file. This file may not even exist once we get the sysl output generated.

pkg/importer/spanner/import.arrai
```diff
@@ -0,0 +1,14 @@
+let [_, importSql, sqlSchema, syslPackage] = //os.args;
```
andrewemeryanz, line 1: Would `appName` be more appropriate than `sqlSchema`?

Taking the following description of parameters from the ticket:
```
-a, --app-name=<APP-NAME> The name of the Sysl app.
Required if the schema doesn't include a CREATE DATABASE statement.
-p, --package=<PACKAGE> The name of the Sysl package.
```

* `appName` is optional. This should overwrite the existing value if provided or return an error if not provided and required.
* `syslPackage` is optional. This shouldn't be added to the output if not provided.

springwiz: yes ok

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,138 @@
+# size returns the size of an attribute
+let size = \length
+    cond {
+        length = 'MAX': '',
+        //seq.has_prefix('0x', length): '(' ++ //seq.trim_prefix('0x', length) ++ ')',
```
orlade-anz, line 5: Nit: in almost all cases, a string template is shorter _and_ easier to read that a string concatenation.

```suggestion
        //seq.has_prefix('0x', length): $`(${//seq.trim_prefix('0x', length)})`,
```

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,138 @@
...
+        length > 0: $`(${length})`,
+        _: '',
```
orlade-anz, line 7: Nit: This line is redundant. If no cond branch matches, the whole thing evals to `{}`.

Same for all other such lines below.

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,138 @@
...
+    };
+
+# attributePatternToken returns the token based on attribute value
+let attributePatternToken = \entity \attr \token
+    cond {
+        (attr.options?:{}): token,
+        (entity.primary_key where //seq.contains(attr.name, .)): token,
+        let fks = (entity.foreign_keys => ((.foreign_keys where .attribute = attr.name) =>
+            .reference_attribute orderby .)) where . != {}; (fks): token,
```
orlade-anz, line 16: Nit: all values except `{}`, `()` and `0` are equivalent to `true` in this context, so `!= {}` is redundant.

```suggestion
            .reference_attribute orderby .)) where .; (fks): token,
```

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,138 @@
...
+        length > 0: $`(${length})`,
+        _: '',
+    };
+
+# attributePatternToken returns the token based on attribute value
+let attributePatternToken = \entity \attr \token
+    cond {
+        (attr.options?:{}): token,
+        (entity.primary_key where //seq.contains(attr.name, .)): token,
+        let fks = (entity.foreign_keys => ((.foreign_keys where .attribute = attr.name) =>
```
orlade-anz, line 15: You do a lot of work to build this data structure, and them merely check whether it is non-empty. Is there a simpler way to write this condition?

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,138 @@
...
+        length = 'MAX': '',
+        //seq.has_prefix('0x', length): '(' ++ //seq.trim_prefix('0x', length) ++ ')',
+        length > 0: $`(${length})`,
+        _: '',
+    };
+
+# attributePatternToken returns the token based on attribute value
+let attributePatternToken = \entity \attr \token
+    cond {
+        (attr.options?:{}): token,
```
orlade-anz, line 13: Nit: these parens around the cond statements are unnecessary. Surprisingly many parens are unnecessary in a arr.ai, and it can often read more naturally without them.

```suggestion
        attr.options?:{}: token,
```

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,138 @@
+# size returns the size of an attribute
+let size = \length
+    cond {
+        length = 'MAX': '',
+        //seq.has_prefix('0x', length): '(' ++ //seq.trim_prefix('0x', length) ++ ')',
+        length > 0: $`(${length})`,
+        _: '',
+    };
+
+# attributePatternToken returns the token based on attribute value
```
orlade-anz, line 10: Nit: as a reader, this is a bit confusing. What is a token? Why would you just want to return the same argument you passed in in one of four cases? Same below.

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,138 @@
...
+        _: '',
+    };
+
+# sortingOrder determines and appends sorting order
+let sortingOrder = \e \attr
+    let pk = //seq.split('(', //seq.join('', (e.primary_key where //seq.contains(attr.name, .)) orderby .));
+    cond {
+        pk count > 1: ', ~' ++ //seq.trim_suffix(')', //str.lower(pk(1))),
+        _: '',
+    };
```
orlade-anz, line 35: Nit: This is just difficult to understand; the code may happen to work, but it's not really expressing its intent, so it's hard to know if it's right.

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,138 @@
...
+        pk & (entity.attributes >> .name): false,
+        _: true,
+    };
+
+# attributePattern generates the pattern constituents depending on the mode
+let attributePattern = \entity \attr \mode
+    cond {
+        (mode = 0 && (entity.primary_key where //seq.contains(attr.name, .))): '~pk' ++ sortingOrder(entity, attr),
+        let fks = (entity.foreign_keys => ((.foreign_keys where .attribute = attr.name) =>
+            .reference_attribute orderby .)) where . != {}; (mode = 1 && fks): '~fk',
```
orlade-anz, line 50: Nit: Having such a large cond expr is hard to read. it's better to move the heavy lifting into the branch like this:
```suggestion
        (mode = 1): 
            let fks = (entity.foreign_keys => ((.foreign_keys where .attribute = attr.name) =>
                .reference_attribute orderby .)) where .;
                cond {fks:  '~fk'}
```

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,138 @@
...
+        (mode = 2 && attr.options):
+            let opt = //seq.split('=',attr.options);
+            opt(0) + '=' + "'" + opt(1) + "'",
```
orlade-anz, line 53:
```suggestion
            $`${opt(0)}='${opt(1)}'`,
```

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,138 @@
...
+                            ${.name} <: ${typeInfo(entity, ., .type ++ size(.length), .array)}${cond {.nullable:'?'}} ${allAttributePatterns(entity, .)}` ::\i:
+                        }`
+                    )
+                    orderby .::\i:
+            }
+            `)}
+        ` orderby .:::}
+    `;
+
+# exports
```
orlade-anz, line 135: Nit: rm, redundant

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,138 @@
...
+        isArray: 'sequence of ' ++ type,
+        _: type,
+    };
+
+# transformModel translates the empty model into sysl file
+let transformModel = \model \package
+    # sysl specification
+    # https://github.com/anz-bank/sysl/blob/master/pkg/sysl/sysl.proto
+    $`
+        ########   THIS IS AUTOGENERATED BY sysl   ########
```
orlade-anz, line 117: Nit: The general Golang convention is:
```suggestion
        // Code generated by sysl. DO NOT EDIT.
```

orlade-anz (changes requested):
WIP comments

pkg/importer/spanner/sysl.arrai
springwiz: It is sysl being generated here.

pkg/importer/spanner/sysl.arrai
springwiz: attributePatternToken checks if the attr passed is included in a foreign key and return a token if included. Token is generally a pattern delimiter which needs to be rendered conditionally.

pkg/importer/spanner/sysl.arrai
springwiz: attributePatternToken checks if the attr passed is included in a foreign key and return a token if included. Token is generally a pattern delimiter which needs to be rendered conditionally.

pkg/importer/spanner/sysl.arrai
springwiz: ok

pkg/importer/spanner/sysl.arrai
springwiz: ok

pkg/importer/spanner/sysl.arrai
springwiz: ok

pkg/importer/spanner/sysl.arrai
springwiz: simplified

pkg/importer/spanner/sysl.arrai
springwiz: ok

pkg/importer/spanner/sysl.arrai
springwiz: ok

pkg/importer/spanner/sysl.arrai
springwiz: ok

pkg/importer/spanner/sysl.arrai
springwiz: ok

pkg/importer/spanner/sysl.arrai
springwiz: ok

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,140 @@
...
+            $`foreign_keys=[${//seq.join(', ',(entity.foreign_keys => \keys $`["constraint:${keys.constraint_name}` ++
+            $`","columns:${//seq.join(', ', (keys.foreign_keys => .attribute) orderby .)}"]`) orderby .)}]`,
+        let relatedIndices = ((model.indexes orderby .) where .@item.table_name = entity.name); mode = 3 && relatedIndices:
+            $`indexes=[${//seq.join(', ', ((model.indexes orderby .) where .@item.table_name = entity.name) >>
+            $`"name:${.table_name}` ++
+            $`","unique:${cond {(.unique = 'UNIQUE'):'true', _:'false'}}` ++
+            $`","null_filtered:${cond {(.nullfiltered = 'NULL_FILTERED'):'true', _:'false'}}` ++
+            $`","key_parts:${//seq.join(', ', .key_part orderby .)}` ++
+            cond {.storing_col: $`","storing:' ${//seq.join(', ',.storing_col)}`}  ++
+            cond {.interleaved_table: $`","interleave_in:' ${.interleaved_table}`})}"]`,
```
orlade-anz, line 84: You don't need to concat strings in a template. If you're worried about line length, try something that more cleanly maps to your intent:

```suggestion
            $`"${[
                $`name:${.table_name}`,
                $`unique:${cond {(.unique = 'UNIQUE'):'true', _:'false'}}`,
                $`null_filtered:${cond {(.nullfiltered = 'NULL_FILTERED'):'true', _:'false'}}`,
                $`key_parts:${.key_part orderby .::, }`,
                .storing_col && $`storing:' ${.storing_col::,}`,
                .interleaved_table && $`interleave_in:' ${.interleaved_table}`,
            ] where .@item::",":"}`
```
Not sure why those `'`s are there in `storing_col` and `interleaved_table`?

Also nit: are `.unique` and `.nullfiltered` ever anything other than `'UNIQUE'` and `'NULL_FILTERED'`? If not, just store a `true` or `false` instead and the `cond` will be simpler (or just use their truthiness like `.unique && 'true' || 'false'`).

orlade-anz (changes requested):

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,140 @@
...
+    let fkAttr = entity.foreign_keys => ((.foreign_keys where .attribute = attr.name) =>
+        .reference_attribute orderby .) where .;
+    let fkTable = entity.foreign_keys => ((.foreign_keys where .attribute = attr.name) =>
+        .reference_table orderby .) where .;
+    cond {
+        fkAttr | fkTable:  //seq.join('', (fkTable => //seq.join('', .) orderby .)) ++ '.'
+            ++ //seq.join('', (fkAttr => //seq.join('', .) orderby .)),
+        isArray: $`sequence of ${type}`,
+        _: type,
+    };
```
orlade-anz, line 115: There are a lot of levels of stuff going on here... hard to follow.

I think this does the same thing, but makes it clearer what the logic is that is acting on the inputs.
```suggestion
let typeInfo = \entity \attr \type \isArray
    let attrFks = entity.foreign_keys => .foreign_keys where .attribute = attr.name;
    let fkAttr = attrFks => .reference_attribute where .;
    let fkTable = attrFks => .reference_table where .;
    # Joins a set of sets into a single alphabetically-ordered string.
    let flatJoin = \sofs $`${sofs => $`${. orderby .::}` orderby .::}`;
    cond {
        fkAttr | fkTable: $`${flatJoin(fkTable)}.${flatJoin(fkAttr)}`,
        isArray: $`sequence of ${type}`,
        _: type,
    };
```

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,140 @@
...
+        pk & (entity.attributes >> .name): false,
+        _: true,
+    };
+
+# attributePattern generates the pattern constituents depending on the mode
+let attributePattern = \entity \attr \mode
+    cond {
+        mode = 0 && (entity.primary_key where //seq.contains(attr.name, .)): $`~pk${sortingOrder(entity, attr)}`,
+        let fks = (entity.foreign_keys => ((.foreign_keys where .attribute = attr.name) => .reference_attribute orderby .)) where .;
+        mode = 1 && fks: '~fk',
```
orlade-anz, line 60: Nit: don't need the `let`, and can short-circuit on the `mode` arg:
```suggestion
        mode = 1 && entity.foreign_keys => (.foreign_keys where .attribute = attr.name) => .reference_attribute where .: '~fk',
```
(untested)

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,140 @@
...
+        }
+    };
+
+# compareColumnOrder compares the col order between primary key and table
+let compareColumnOrder = \entity
+    let pk = (entity.primary_key orderby .) >> //seq.split('(', .) >> .(0);
+    cond {
+        pk & (entity.attributes >> .name): false,
+        _: true,
+    };
```
orlade-anz, line 53: Nit: if what you're testing here is that `pk` is not a subset of `entity.attributes >> .name` (i.e. that there's no overlap), you can just use the (not) subset operator:
```suggestion
    pk !(<=) (entity.attributes >> .name);
```

Also be aware that all the set operations, including set intersection (`&`), do very literal checks. `entity.attributes >> .name` is an array, so `pk & (entity.attributes >> .name)` will only match if the indexes in both arrays match. It's hard to follow the logic because of the structure of entity, but it looks like there's a risk of missing matches based on the index. If that's actually intentional, a comment would help to clarify.

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,140 @@
...
+
+# sortingOrder determines and appends sorting order
+let sortingOrder = \e \attr
+    let keyOrder = //seq.split('(', //seq.join('', (e.primary_key where //seq.contains(attr.name, .)) orderby .));
+    cond {
+        keyOrder count > 1: cond keyOrder(1) {
+            'asc)': ', ~asc',
+            'desc)': ', ~desc'
+        }
+    };
```
orlade-anz, line 45: Rather than doing this strange split and double check, can you better express what you're actually doing? It looks like you're building a big string, then checking whether it contains `(asc)` or `(desc)`, or possibly checking which one comes first.

Perhaps:
```suggestion
    let keys = //seq.join('', e.primary_key where //seq.contains(attr.name, .) orderby .);
    cond //re.compile('\\((asc|desc)\\)').match(keys) {[[_, ord], ...]: ord rank (:.@)};
```

orlade-anz (changes requested):

pkg/importer/spanner/sysl.arrai
springwiz: Intent looks pretty clear from the code get the chosen order from primary_key attribute and return an '~asc' or `~desc` depending on the chosen orderby.

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,140 @@
...
+    cond {
+        attr.options?:{}: true,
+        (entity.primary_key where //seq.contains(attr.name, .)): true,
+        let fks = (entity.foreign_keys => ((.foreign_keys where .attribute = attr.name) =>
+            .reference_attribute orderby .)) where .; fks: true,
+        attr.length = 'MAX': true,
+        //seq.has_prefix('0x', attr.length): true,
+    };
+
+# hasEntityPattern checks whether an entity has associated patterns
```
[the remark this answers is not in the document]
springwiz: So patterns is anything included within [], you could have them associated with either entity or attribute.

pkg/importer/importer.go
```diff
@@ -45,6 +46,9 @@ func Factory(filePath string, file []byte, logger *logrus.Logger) (Importer, err
 	case Grammar.Name:
 		logger.Debugln("Detected Grammar file")
 		return nil, fmt.Errorf("importer disabled for: %s", fileType.Name)
+	case SpannerSQL.Name:
+		logger.Debugln("Detected Spanner sql file")
```
orlade-anz, line 50: Nit
```suggestion
		logger.Debugln("Detected Spanner SQL file")
```

pkg/importer/importer_test.go
```diff
@@ -103,6 +104,42 @@ func TestLoadXSDFromTestFiles(t *testing.T) {
...
+		return "", err
+	}
+	bytes, err := ioutil.ReadFile(absFilePath)
+	return string(bytes), err
+}
+
+func TestCompareSpannerSQLWithGolden(t *testing.T) {
+	// Load generated sysl module
+	genSysl, err := generateSyslFromSpannerSQL()
+	require.Nil(t, err)
```
orlade-anz, line 133: Nit: `require.Nil` is odd. AFAICT we're using `testify` elsewhere, so you probably want `assert.NoError(t, err)`.

pkg/importer/importer_test.go
```diff
@@ -103,6 +104,42 @@ func TestLoadXSDFromTestFiles(t *testing.T) {
...
+
+	// Load golden sysl module
+	goldenSysl, err := readGoldenSyslForSpannerSQL()
+	require.Nil(t, err)
+
+	isValid := reflect.DeepEqual(genSysl, goldenSysl)
+	require.True(t, isValid)
```
orlade-anz, line 140: If you're just comparing strings, you don't need deep equal, just `assert.Equal`:
```suggestion
	assert.Equal(t, goldenSysl, genSysl)
```

pkg/importer/spanner/import.arrai
```diff
@@ -0,0 +1,18 @@
...
+    let model = spanner.applyStmt(stmts, spanner.emptyModel);
+    $'${sysl.transformModel(
+        cond {
+            (model.schema): model,
+            _: (
+                entities: model.entities,
+                indexes: model.indexes,
+                schema: {(name: appName)},
+            )
+        }, syslPackage)}';
```
orlade-anz, line 14: `sysl.transformModel` is already a string.
```suggestion
    sysl.transformModel(
        cond {
            (model.schema): model,
            _: (
                entities: model.entities,
                indexes: model.indexes,
                schema: {(name: appName)},
            )
        }, syslPackage);
```

pkg/importer/spanner/import_test.arrai
```diff
@@ -0,0 +1,3 @@
+let importlib = //{./import};
```
orlade-anz, line 1: Files ending in `_test.arrai` will be treated as unit test specs by `arrai test`. They should not expect any `//os.args`, and should return only `true` or a data structure full of `true` values.

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,141 @@
...
+                !table ${entity.name} ${allEntityPatterns(entity, model)}:
+                    ${entity.attributes >>
+                        $`
+                            ${.name} <: ${typeInfo(entity, ., .type ++ size(.length), .array)}${cond {.nullable:'?'}} ${allAttributePatterns(entity, .)}` ::\i:
+                        }`
+                    )
+                    orderby .::\i:
+            }
+            `)}
+        ` orderby .:::}
```
orlade-anz, line 137: Nit: you only need `::` unless you're planning to append something to the end.

```suggestion
        ` orderby .::}
```

orlade-anz (changes requested):

pkg/importer/importer_test.go
springwiz: Not sure if there is a difference, a lot of the code still uses require.Nil. anyways a small change.

pkg/importer/importer_test.go
springwiz: ok

pkg/importer/spanner/import_test.arrai
springwiz: will rename

pkg/importer/importer_test.go
orlade-anz: Sorry yeah, I just looked it up and require is probably right in this instance. Both require and assert are part of the same library, but `require` terminates immediately if it fails, whereas `assert` can allow execution to continue to collect more errors.

`require` is probably right in this instance.

pkg/importer/importer.go
springwiz: ok

springwiz: ok

springwiz: ok

springwiz: ok

pkg/importer/spanner/sysl.arrai
orlade-anz: The intent of splitting on `(` is unclear; it requires knowledge of the structure of `primary_key` (fair enough), and the fact that the sort order is encoded into the `primary_key` column name string (which it shouldn't be), and confidence in the fact that the only time `(` will exist in a column name is immediately before a sort order (which is fragile if this assumption ever changes).

For example: what if more data is encoded in the name? Or what if there are multiple columns with similar names and different sort orders? Your split for `['foo(asc)', 'foobar(desc)']` will return something like `['foo', 'asc)foobar', 'desc)']`, and keyOrder(1) will be `asc)foobar` which won't match.

What I mean by "intent" is that the implementation should closely match the conceptual logic of the function. In this case, the function is conceptually searching for either `(asc)` or `(desc)` in one of the PK column name strings (not clear which one if there are multiple). When you look at the implementation it's joining strings, splitting on `(` and picking the second element. None of that is related to the concept of searching for sort orders.

Having looked into the structure of `primary_key`, I think my suggestion could be further improved. It should literally say "check each column for an order, and return that order with a tilde in front." `//seq.contains(attr.name)` is not appropriate in this case, since it's supposed to be an exact match.

```
    let re = //re.compile($`${attr.name}\((asc|desc)\)`);
    let order = e.primary_key => (re.match(.)(0)?(1)?:{} rank (:.@)) where .;
    cond order {{x}: $`~${x}`}
```

An even better solution would be to not embed the sort order in the column name; make it a tuple with `name` and `order` instead of a string.

pkg/importer/spanner/sysl.arrai
springwiz: fine

pkg/importer/spanner/import.arrai
springwiz: fine

pkg/importer/spanner/sysl.arrai
springwiz: fine

pkg/importer/spanner/sysl.arrai
springwiz: Its comparing the col order between table and primary key, there is a comment already.

pkg/importer/spanner/sysl.arrai
springwiz: let flatJoin = \sofs $`${sofs => $`${. orderby .::}` orderby .::}`; breaks with a missing close. But all the { seems to be in place.

pkg/importer/spanner/sysl.arrai
springwiz: ok

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,140 @@
...
+let entityPattern = \entity \model \mode
+    cond {
+        mode = 0 && entity.primary_key count > 1 && compareColumnOrder(entity):
+            $`primary_key="${//seq.join(', ', entity.primary_key orderby .)}"`,
+        mode = 1 && entity.cluster:
+            //seq.join('', entity.cluster >> $`interleave_in_parent="${.interleaved_in}", interleave_on_delete="${.on_delete}"`),
+        mode = 2 && entity.foreign_keys:
+            $`foreign_keys=[${//seq.join(', ',(entity.foreign_keys => \keys $`["constraint:${keys.constraint_name}` ++
+            $`","columns:${//seq.join(', ', (keys.foreign_keys => .attribute) orderby .)}"]`) orderby .)}]`,
+        let relatedIndices = ((model.indexes orderby .) where .@item.table_name = entity.name); mode = 3 && relatedIndices:
```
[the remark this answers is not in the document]
springwiz: ok

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,128 @@
...
+    let fk = cond {
+        entity.foreign_keys:
+            $`foreign_keys=[${//seq.join(', ',(entity.foreign_keys =>
+                \keys $`["constraint:${keys.constraint_name}","columns:${//seq.join(', ', (keys.foreign_keys => .attribute) orderby .)}"]`) orderby .)}]`,
+    };
+    let indices = cond {
+        (model.indexes where .table_name = entity.name):
+            $`indexes=[${//seq.join(', ', ((model.indexes orderby .) where .@item.table_name = entity.name) >>
+                $`"name:${.table_name}","unique:${.unique}","null_filtered:${.nullfiltered}","key_parts:${//seq.join(', ', .key_part orderby .)}${cond {.storing_col: $`","storing:${//seq.join(', ',.storing_col)}`}}${cond {.interleaved_table: $`","interleave_in:${.interleaved_table}`}}`)}"]`,
+    };
```
orlade-anz, line 76: This is quite long and a mix of operations. My first pass at making it simpler was:
```
    let indices = cond model.indexes where .table_name = entity.name {
        indexes: $`
            indexes=[${indexes => $`
                "name:${.table_name}", "unique:${.unique}", "null_filtered:${.nullfiltered}", "key_parts:${.key_part orderby .::, )
                }${cond {.storing_col: $`", "storing:${.storing_col::, }`}
                }${cond {.interleaved_table: $`", "interleave_in:${.interleaved_table}`}}
            ` orderby .::, }"]
        `,
    };
```

But inspired by your array construction previously, this is probably even nicer:

```suggestion
    let indices = cond model.indexes where .table_name = entity.name {indexes:
        let indexStrs = indexes => ([
            $`name:${.table_name}`,
            $`unique:${.unique}`,
            $`null_filtered:${.nullfiltered}`,
            $`key_parts:${.key_part orderby .::, }`,
            cond {.storing_col: $`storing:${.storing_col::, }`},
            cond {.interleaved_table: $`"interleave_in:${.interleaved_table}`},
        ] where .@item -> $`"${.::", "}"`);
        $`indexes=[${indexStrs orderby .::, }]
```

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,128 @@
...
+    //seq.join(', ', ([pk, cluster, fk, indices] where .@item != ''));
```
orlade-anz, line 77: Nit: again, I think it would be cleaner overall to have this function just produce an array of pattern strings, and leaves the joining to whatever function adds the `[...]` brackets around the result (i.e. `appendEntityPatterns`).
```suggestion
    [pk, cluster, fk, indices] where .@item != '';
```

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,128 @@
...
+
+# appendEntityPatterns appends the entity patterns together.
+let appendEntityPatterns = \entity \model
+    cond {
+        hasEntityPattern(entity, model): $`[${entityPatterns(entity, model)}]`
```
orlade-anz, line 82:
```suggestion
        hasEntityPattern(entity, model): $`[${entityPatterns(entity, model)::, }]`
```

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,128 @@
...
+    };
+
+# appendAttributePatterns appends the attribute patterns together.
+let appendAttributePatterns = \entity \attr
+    cond {
+        hasAttributePattern(entity,attr): $`[${attributePatterns(entity, attr)}]`
```
orlade-anz, line 88:
```suggestion
        hasAttributePattern(entity,attr): $`[${attributePatterns(entity, attr)::, }]`
```

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,128 @@
...
+        .reference_attribute orderby .)) where .:
+            '~fk'
+    };
+    let length = cond {
+        attr.length = 'MAX': '~max',
+    };
+    let hexPrefix = cond {
+        //seq.has_prefix('0x', attr.length): '~hex',
+    };
+    //seq.join(', ', ([options, pk, fk, length, hexPrefix] where .@item != ''));
```
orlade-anz, line 54: See comment below on return from `entityPatterns` (sorry for being out of order)
```suggestion
    [options, pk, fk, length, hexPrefix] where .@item != '';
```

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,128 @@
...
+    let indices = cond {
+        (model.indexes where .table_name = entity.name):
+            $`indexes=[${//seq.join(', ', ((model.indexes orderby .) where .@item.table_name = entity.name) >>
+                $`"name:${.table_name}","unique:${.unique}","null_filtered:${.nullfiltered}","key_parts:${//seq.join(', ', .key_part orderby .)}${cond {.storing_col: $`","storing:${//seq.join(', ',.storing_col)}`}}${cond {.interleaved_table: $`","interleave_in:${.interleaved_table}`}}`)}"]`,
+    };
+    //seq.join(', ', ([pk, cluster, fk, indices] where .@item != ''));
+
+# appendEntityPatterns appends the entity patterns together.
+let appendEntityPatterns = \entity \model
+    cond {
```
orlade-anz, line 81: Nit: If the body of the cond is a one-liner, probably just make the whole cond a one-liner.

pkg/importer/spanner/sysl.arrai
orlade-anz, line 88: Nit: space
```suggestion
        hasAttributePattern(entity, attr): $`[${attributePatterns(entity, attr)}]`
```

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,128 @@
...
+
+# typeInfo generates the type info for an attribute.
+let typeInfo = \entity \attr \type \isArray
+    let fkAttr = entity.foreign_keys => ((.foreign_keys where .attribute = attr.name) =>
+        .reference_attribute orderby .) where .;
+    let fkTable = entity.foreign_keys => ((.foreign_keys where .attribute = attr.name) =>
+        .reference_table orderby .) where .;
+    cond {
+        fkAttr | fkTable:  //seq.join('', (fkTable => //seq.join('', .) orderby .)) ++ '.'
+            ++ //seq.join('', (fkAttr => //seq.join('', .) orderby .)),
```
orlade-anz, line 99: Upon closer inspection, this seems to require that both `fkAttr` and `fkTable` are non-empty? In that case `fkAttr && fkTable` would be a better condition.

Also the `//seq.join('', ...)` calls are adding a lot of noise. Perhaps `flatJoin` wasn't the best idea, but it would still be nice to get rid of them and focus more on the data you're operating on. Maybe:
```suggestion
        fkAttr && fkTable: $`${fkTable => $`${.::}` orderby .::}.${fkAttr => $`${.::}` orderby .::}`,
```

Or if you really don't like the `${.::}` pattern, consider using currying to do something like `let join = //seq.join('')`, with which you could just do a simpler:

```
$`${join(fkTable => join(.) orderby .)}.${join(fkAttr => join(.) orderby .)}`
```

pkg/importer/spanner/sysl.arrai
orlade-anz: My point is not about the language, but the style of the comment. I realise now that files generated by `sysl import` have their own style of header comment.

However `pkg/importer/writer.go` adds this header to the beginning of the generated source, so if this arr.ai script is being used within the context of `sysl import`, then this script shouldn't be adding a header comment like this at all.

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,128 @@
...
+            #>>>
+                !table ${entity.name} ${appendEntityPatterns(entity, model)}:
+                    ${entity.attributes >>
+                        $`
+                            ${.name} <: ${typeInfo(entity, ., .type ++ size(.length), .array)}${cond {.nullable:'?'}} ${appendAttributePatterns(entity, .)}` ::\i:
+                        }`
+                    )
+                    orderby .::\i:
+            }
+            `)}
```
orlade-anz, line 123: You shouldn't need this magic for indenting. Something like the following should work (thanks to Novan's fixes), or there's a bug:
```suggestion
                ${(model.entities => \entity $`
                    !table ${entity.name} ${appendEntityPatterns(entity, model)}:
                        ${entity.attributes >> $`
                            ${.name} <: ${typeInfo(entity, ., .type ++ size(.length), .array)}${cond {.nullable:'?'}} ${appendAttributePatterns(entity, .)}
                        `::\i:}
                `) orderby .::\i:}
```

orlade-anz (changes requested):
Looking much nicer! More comments for extra polish :)

pkg/importer/spanner/sysl.arrai
springwiz: This doesn't use pkg/importer/writer.go to write the sysl file.

springwiz: Doesn't work even in the lastest arrai ver 158 without hack.

springwiz: comment fixed as well.

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,129 @@
...
+# appendEntityPatterns returns a pattern for the entity patterns together.
+let appendEntityPatterns = \entity \model cond { hasEntityPattern(entity, model): $`[${entityPatterns(entity, model)}]`};
+
+# appendAttributePatterns returns a pattern for the attribute patterns together.
+let appendAttributePatterns = \entity \attr cond { hasAttributePattern(entity, attr): $`[${attributePatterns(entity, attr)}]`};
+
+# typeInfo generates the type info for an attribute.
+let typeInfo = \entity \attr \type \isArray
+    let fkAttr = entity.foreign_keys => ((.foreign_keys where .attribute = attr.name) =>
+        .reference_attribute orderby .) where .;
```
orlade-anz, line 92: Nit: I've seen `entity.foreign_keys => ((.foreign_keys where .attribute = attr.name) => .reference_attribute orderby .) where .;` four times in this file, and it makes whatever code it's in look more complex. Candidate for refactoring into a function?

orlade-anz (changes requested):
Looking good! Down to nits at this point :)

pkg/importer/spanner/sysl.arrai
orlade-anz: Other importers use `pkg/importer/writer.Write()` to write out their Sysl string, and `Write` prepends this autogenerated header. Ideally use that for consistency (export `writeHeader` if necessary).

pkg/importer/spanner/sysl.arrai
```diff
@@ -0,0 +1,129 @@
...
+# sortingOrder determines and appends sorting order.
+let sortingOrder = \e \attr
+    let re = //re.compile($`${attr.name}\((asc|desc)\)`);
+    let keyOrder = e.primary_key => (re.match(.)(0)?(1)?:{} rank (:.@)) where .;
+    cond keyOrder {{x}: $`~${x}`};
+
+# compareColumnOrder compares the col order between primary key and table.
+let compareColumnOrder = \entity
+    let pk = (entity.primary_key orderby .) >> //seq.split('(', .) >> .(0);
+    pk !(<=) (entity.attributes >> .name);
```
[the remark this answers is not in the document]
springwiz: That breaks the working code.

pkg/importer/spanner/sysl.arrai
springwiz: refactored

pkg/importer/spanner/sysl.arrai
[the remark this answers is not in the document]
orlade-anz: In what way? Is the order in which the attributes are defined in `entity` relevant to the behaviour of this function? Or are they guaranteed to be alphabetical? If not, then the code's reliance on them is confusing, and potentially a bug.

My main issue really is that I can't tell from reading either code or comment what the expected behaviour of this code is. I'd like to try to suggest an edge case counterexample to demonstrate a failure mode, but I can't without first understanding that intention.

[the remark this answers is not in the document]
springwiz: This compares on order of the columns in the primary key and the order in which they are specified in the table(entity.attributes is a tuple) and that's why I convert them to an array and compare the order. Your first statement randomises the order by converting it to a set and that's why it fails.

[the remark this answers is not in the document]
springwiz: The comment on the method is # compareColumnOrder compares the col order between primary key and table. which explains that this method is order sensitive.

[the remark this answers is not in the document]
orlade-anz: Following up offline conversation, `entity.primary_key` should be sorted by order of definition, not alphabetical. Addressed through use of string offset.

orlade-anz:
Some notes from real usage (may address in followup PRs):

- `!table`s without attributes shouldn't have an empty ` []` after their name.
- Add newline between autogenerated header and first line of output.
- There should be no trailing whitespace in the generated file. There is currently some on almost every line.
- The generated `spanner.arrai.go` must be re-generated on `make install`, probably in a step called `generate` as a pre-requisite of `build`.
- Ideally the Go generation script shouldn't assume a working directory, since everything it needs is relative to itself. It can start with the following to ensure the working directory is its own:
```bash
SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
cd $SCRIPT_DIR
```

pkg/importer/spanner/spanner.arrai
```diff
@@ -0,0 +1,329 @@
...
+                     | ADD_CONSTRAINT  = ("ADD" table_constraint)
+                     | DROP_CONSTRAINT = ("DROP CONSTRAINT" constraint_name);
+attr_alteration     -> "ALTER COLUMN" column_name (attr_type | "SET" options_def);
+
+drop_table          -> "DROP TABLE" table_name;
+
+drop_index          -> "DROP INDEX" index_name;
+
+options_def         -> "OPTIONS ("(option):",",? ")";
+option              -> "allow_commit_timestamp="("true"|"false");
```
orlade-anz, line 37: Support more whitespace.
```suggestion
options_def         -> "OPTIONS" "("(option):",",? ")";
option              -> "allow_commit_timestamp" "=" ("true"|"false");
```

orlade-anz (changes requested):

orlade-anz:
A trick for trimming trailing whitespace:

```arrai
let sysl = $`...`;

# Trim trailing whitespace.
let re = //re.compile(`^(.*[^\s])\s*$`);
$`${//seq.split('\n', sysl) >> cond re.match(.) {[[_, s]]: s, _: ''}::\n}`
```

pkg/importer/spanner/sysl.arrai
orlade-anz: Here's some similar code I wrote to generate Sysl from some XLSX data. The indenting works just fine with the latest version of arr.ai (though there are more 

```arrai
let sysl = $`
App[~db]:
    ${tables orderby .@row >> $`
        !table ${.table}:
            ${(.fields orderby .@row) >> $`
                ${.field}${cond {.data_type: ' ' + $`<: ${.data_type}`}}${
                    cond {.description: $`
                        :
                            @description =:
                                ${//seq.split('\n', .description) >> $`| ${.}`::\i}
                    `}
                }
            `::\i}
    `::\i\n}
`;
```

Note the workaround below was due to a bug with string templates eating leading whitespace too aggressively, but that's a small and different problem.
```
' ' + $``
```

PR #997, opened 2020-08-10
Jack/add one of openapi

cuminandpaprika:
Adds very limited preliminary support for the `OneOf` syntax in the OpenAPI importer.

Currently only supports OpenAPI OneOf syntax where each item is a $ref.

e.g
```yaml
Pet:
        type: object
        oneOf:
          - $ref: '#/components/schemas/Cat'
          - $ref: '#/components/schemas/Dog'
```

Not currently supported:

The YAML below will produce a syntax error in the output Sysl
```
CatDog: 
     type: object
     oneOf:
       - type: string
          format: email
       - type: string
          format: uuid
```
Partial fix for https://github.com/anz-bank/sysl-go/issues/95

Checklist:
- [x] Added related tests


pkg/importer/writer_test.go
```diff
@@ -0,0 +1,30 @@
...
+		name     string
+		arg      Type
+		expected string
+	}{
+		{"writeUnion",
+			&Union{name: "TestUnion", Options: FieldList{{Name: "Apple"}, {Name: "Orange"}}},
+			"!union TestUnion:\n    Apple\n    Orange\n",
+		},
+	}
+	for _, tt := range tests {
```
orlade-anz, line 22: Nit: Does this really need an array and a range?

cuminandpaprika: Definitely not necessary at the moment, but planning to add a bunch more test cases in a future PR 👍

PR #1000, opened 2020-08-11
Add rough util test

orlade-anz:
Example of an arr.ai unit test specification for the arr.ai demo code.


orlade-anz:
Added separately.

PR #1002, opened 2020-08-12
fix sysl modules cache path

ChloePlanet:
- fixed github file content commit sha
- added field `version` to `SourceContext` and passed `mod.version` in `SourceContext`


pkg/mod/github.go
```diff
@@ -69,8 +69,12 @@ func (d *githubMgr) Get(filename, ver string, m *Modules) (*Module, error) {
 	if err != nil {
 		return nil, err
 	}
-	if ver == "" {
-		ver = "v0.0.0-" + fileContent.GetSHA()[:12]
+	if ver == "" || ver == MasterBranch {
+		ref, _, err := d.client.Git.GetRef(ctx.Background(), repoPath.owner, repoPath.repo, "heads/"+MasterBranch)
```
cuminandpaprika, line 73:
```suggestion
		ref, _, err := d.client.Git.GetRef(context.Background(), repoPath.owner, repoPath.repo, "heads/"+MasterBranch)
```

pkg/parse/fs.go
```diff
@@ -26,8 +27,8 @@ type fsFileStream struct {
 	filename string
 }
 
-func newFSFileStream(filename string, fs afero.Fs) (*fsFileStream, error) {
-	f, err := fs.Open(filename)
+func newFSFileStream(filename string, fs afero.Fs) (*fsFileStream, *mod.Module, error) {
+	f, mod, err := fs.OpenWithModule(filename)
 	if err != nil {
 		return nil, err
```
cuminandpaprika, line 33:
```suggestion
		return nil,nil, err
```

pkg/parse/fs.go
```diff
@@ -38,5 +39,5 @@ func newFSFileStream(filename string, fs afero.Fs) (*fsFileStream, error) {
...
```
cuminandpaprika, line 39:
```suggestion
		return nil,nil, err
```

pkg/parse/fs.go
```diff
@@ -26,8 +27,8 @@ type fsFileStream struct {
 	filename string
 }
 
-func newFSFileStream(filename string, fs afero.Fs) (*fsFileStream, error) {
-	f, err := fs.Open(filename)
+func newFSFileStream(filename string, fs afero.Fs) (*fsFileStream, *mod.Module, error) {
```
cuminandpaprika, line 30:
```suggestion
func newFSFileStream(filename string, fs mod.Fs) (*fsFileStream, *mod.Module, error) {
```

cuminandpaprika: Actually I think it might be better to create a new Fs interface inside of the mod package, that embeds the afero.Fs interface

pkg/sysl/sysl.proto
```diff
@@ -21,6 +21,7 @@ message SourceContext {
     string      text    = 4;
 
     Delta   delta       = 6;
+    string  version     = 7;
```
cuminandpaprika, line 24: We should probably add a comment here describing what this field is used for
```suggestion
    string  version     = 7; // This is populated when Github Mode is used for sysl modules, e.g v0.0.0-c63b9e92813a where v0.0.0 is the version tag and c63b9e92813a is the first 12 chars of the commit hash
```

pkg/mod/github.go
```diff
@@ -69,8 +69,12 @@ func (d *githubMgr) Get(filename, ver string, m *Modules) (*Module, error) {
...
 		return nil, err
 	}
-	if ver == "" {
-		ver = "v0.0.0-" + fileContent.GetSHA()[:12]
+	if ver == "" || ver == MasterBranch {
+		ref, _, err := d.client.Git.GetRef(ctx, repoPath.owner, repoPath.repo, "heads/"+MasterBranch)
+		if err != nil {
+			return nil, err
+		}
+		ver = "v0.0.0-" + ref.GetObject().GetSHA()[:12]
```
cuminandpaprika, line 77: Also short commit hashes are usually 7 digits (didn't know that off the top of my head) https://www.shellhacks.com/git-get-short-hash-sha-1-from-long-hash-head-log/

Is there any reason why we're grabbing the first 12?

ChloePlanet: Cuz go modules use first 12

cuminandpaprika: Ahhh I see! No worries sounds good

PR #1024, opened 2020-08-20
Remove explicit config of default golangci-lint version

orlade-anz:
* Please attach **WIP** tag when this PR is still work in progress.
* Please attach **breaking-change** tag if this PR potentially causes other components to fail or changes previous sysl executable's behaviour.

Fixes # .

- 
- 

Checklist:
- [ ] Added related tests

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -154,5 +154,7 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
 
-      - name: Run golangci-lint
+      - name: Run golangci-lint 1.30.0
         uses: actions-contrib/golangci-lint@v1
+        with:
+          golangci_lint_version: 1.30.0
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -154,7 +154,5 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
 
-      - name: Run golangci-lint 1.30.0
+      - name: Run golangci-lint
         uses: actions-contrib/golangci-lint@v1
-        with:
-          golangci_lint_version: 1.30.0
```
cuminandpaprika, line 160: So, without passing in the golangci_lint_version, it'll default to v1.23.0. Is that the intended behaviour? 

https://github.com/actions-contrib/golangci-lint/blob/master/Dockerfile

orlade-anz: Removed from the right file instead :)

cuminandpaprika (approved):
LGTM!

PR #1025, opened 2020-08-24
Remove code generate examples to resolve cyclic dependency

orlade-anz:
There's currently a weird dependency from `sysl` to `sysl-go`, pulled in by the code gen examples that contain `sysl-go` source code. This introduces a dependency cycle and messes up tooling that performs dependency analysis.

This is a proposal to remove those examples, presumably a precursor to remove all examples and migrating them to `sysl-website` or similar.


joshcarp (approved):

ChloePlanet (approved):
note: some tests failed

PR #1042, opened 2020-09-02
Updated int and float to add new attr bitWidth.

ericzhang6222:
* Please attach **WIP** tag when this PR is still work in progress.
* Please attach **breaking-change** tag if this PR potentially causes other components to fail or changes previous sysl executable's behaviour.

Fixes # .

- 
- 

Checklist:
- [ ] Added related tests


pkg/parse/tests/numbers_format.sysl
```diff
@@ -0,0 +1,20 @@
+Numbers:
+    !type A:
+        aAttr0 <: int
+        aAttr1 <: float
+
+    !type B:
+        bAttr0 <: int32
+        bAttr1 <: int32(4)
```
andrewemeryanz, line 8: Only include the following types in your test:
```
int
int32
int64
float
float32
float64
```

Types like `int32(4)` currently parse but are unsupported. I've [added](https://github.com/anz-bank/sysl/issues/1044) a ticket to deprecate unsupported values.

pkg/parse/utils.go
```diff
@@ -54,6 +54,18 @@ func (s sourceCtxHelper) get(start, end antlr.Token, withText bool) *sysl.Source
 	}
 }
 
+func getBitWidth(t *sysl.Type) int32 {
```
andrewemeryanz, line 57: Document:
```
// getBitWidth returns the bit width of the type, or 0 if no such constraint exists.
```

pkg/parse/utils.go
```diff
@@ -54,6 +54,18 @@ func (s sourceCtxHelper) get(start, end antlr.Token, withText bool) *sysl.Source
...
+	if t.Constraint != nil {
+		for _, item := range t.Constraint {
+			if item != nil && item.BitWidth > 0 { // 32 or 64
```
andrewemeryanz, line 60: Remove `32 or 64`. It may be any arbitrary positive integer in the future.

andrewemeryanz (changes requested):
Minor changes requested.

pkg/sysl/sysl.proto
```diff
@@ -295,6 +295,7 @@ message Type {
         Resolution resolution = 3;
         int32 precision = 4;
         int32 scale = 5;
+        int32 bitWidth = 6;
```
andrewemeryanz, line 298: Sorry. My error here. This should be named `bit_width`.

andrewemeryanz (changes requested):

andrewemeryanz (approved):

andrewemeryanz (approved):

PR #1065, opened 2021-05-12
Encrypting and testing target data

ashwinsajiv:
Test for encrypting data on CI and eventually target data.


ashwinsajiv:
Mission accomplished ...

